PR #3, opened 2025-10-28
Delete toggle subtask

winnwy:


blotztask-api/Modules/Tasks/Controllers/SubTaskController.cs
```diff
@@ -50,4 +49,16 @@ public async Task<IActionResult> ReplaceSubtasks(
...
 
+    [HttpDelete("subtasks/{subtaskId}")]
+    public async Task<IActionResult> DeleteSubtask(int subtaskId, CancellationToken ct)
+    {
+        var command = new DeleteSubtaskCommand { SubtaskId = subtaskId };
+        var deleted = await deleteSubtaskCommandHandler.Handle(command, ct);
+        if (deleted == null)
+        {
+            return NotFound("Subtask not found");
+        }
```
sol-wizard: why the check is in controller layer?

blotztask-api/Modules/Tasks/Commands/SubTasks/DeleteSubtask.cs
```diff
@@ -0,0 +1,27 @@
...
+
+public class DeleteSubtaskCommandHandler(BlotzTaskDbContext db, ILogger<DeleteSubtaskCommandHandler> logger)
+{
+    public async Task<string?> Handle(DeleteSubtaskCommand command, CancellationToken ct = default)
+    {
+        var subtaskToDelete = await db.Subtasks.FindAsync(command.SubtaskId, ct);
+        if (subtaskToDelete == null)
+        {
+            return null;
+        }
```
sol-wizard, line 21: if not found you should throw error not return null

sol-wizard (changes requested):

blotztask-api/Modules/Tasks/Domain/Entities/Subtask.cs
```diff
@@ -7,9 +7,9 @@ public class Subtask
     public required string Title { get; set; }
     public string? Description { get; set; }
     public TimeSpan? Duration { get; set; }
-    public int Order { get; init; }
+    public int Order { get; set; }
     public bool IsDone { get; set; }
-    public DateTime CreatedAt { get; init; }     
+    public DateTime CreatedAt { get; init; }
```
sol-wizard, line 12: you changed entity, do you do a migrations?

sol-wizard (changes requested):

sol-wizard (approved):

PR #37, opened 2025-11-21
feat: add dedicated endpoint for subtask status updates

Emilia-Z:


blotztask-api/Modules/Tasks/Commands/SubTasks/UpdateSubtaskStatus.cs
```diff
@@ -0,0 +1,49 @@
...
+using BlotzTask.Shared.Exceptions;
+using Microsoft.EntityFrameworkCore;
+
+namespace BlotzTask.Modules.Tasks.Commands.SubTasks;
+
+
+
+public class UpdateSubtaskStatusCommand
+{
+    [Required]
```
nxn-nicole: looks like no need to use required here

blotztask-api/Modules/Tasks/Commands/SubTasks/UpdateSubtaskStatus.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            Id = subtask.Id,
+            UpdatedAt = subtask.UpdatedAt,
+            Message = subtask.IsDone? "Subtask marked as completed" :  "Subtask marked as incompleted"
+        };
+
+    }
+}
+
+public class SubtaskStatusResultDto
```
nxn-nicole: look like frontend doesn't need these information? maybe we don't need to return it

blotztask-api/Modules/Tasks/Commands/SubTasks/UpdateSubtaskStatus.cs
```diff
@@ -0,0 +1,49 @@
+using System.ComponentModel.DataAnnotations;
+using BlotzTask.Infrastructure.Data;
+using BlotzTask.Shared.Exceptions;
+using Microsoft.EntityFrameworkCore;
+
+namespace BlotzTask.Modules.Tasks.Commands.SubTasks;
+
+
+
+public class UpdateSubtaskStatusCommand
```
nxn-nicole: why need an extra command here and doesn't use subtaskId directly?

nxn-nicole (changes requested):

blotztask-api/Modules/Tasks/Commands/SubTasks/UpdateSubtaskStatus.cs
```diff
@@ -0,0 +1,38 @@
+using System.ComponentModel.DataAnnotations;
+using BlotzTask.Infrastructure.Data;
+using BlotzTask.Shared.Exceptions;
+using Microsoft.EntityFrameworkCore;
+
+namespace BlotzTask.Modules.Tasks.Commands.SubTasks;
+
+
+
+public class @int
```
nxn-nicole: why need to use @int? I feel like you use Handle(int subtaskId, below is fine

nxn-nicole (changes requested):

nxn-nicole (approved):

PR #44, opened 2025-11-25
fix green spot bug (floating title)

nxn-nicole:


blotztask-api/Modules/Tasks/Queries/Tasks/GetAllTasks.cs
```diff
@@ -17,7 +17,7 @@ public async Task<List<AllTaskItemDto>> Handle(GetAllTasksQuery query, Cancellat
 
         var tasks = await db.TaskItems
             .Where(t => t.UserId == query.UserId)
-            .OrderByDescending(t => t.StartTime ?? DateTimeOffset.MinValue)
+            .OrderByDescending(t => t.StartTime).ThenBy(t => t.Title)
```
sol-wizard, line 20: what if task with no start time ? will this crash or any error ?

sol-wizard (approved):

PR #47, opened 2025-11-25
Long press delete, short press direct edit page

LiChenyang-GZ:


blotztask-mobile/src/app/(protected)/star-spark.tsx
```diff
@@ -41,10 +69,28 @@ export default function StarSparkScreen() {
...
       </View>
+
       {isLoading && <LoadingScreen />}
       {!isLoading && floatingTasks && floatingTasks.length > 0 && (
-        <FloatingTaskDualView tasks={floatingTasks} />
+        <FloatingTaskDualView
+          tasks={floatingTasks}
+          onDeleteTask={handleDeleteTask}
+          deletingId={deletingId}
+          onPressTask={handlePressTask}
```
nxn-nicole, line 74: you can use handlePressTask inside task card then maybe don't need to pass it from the parent component?

blotztask-mobile/src/app/(protected)/star-spark.tsx
```diff
@@ -41,10 +69,28 @@ export default function StarSparkScreen() {
...
         />
       </View>
+
       {isLoading && <LoadingScreen />}
       {!isLoading && floatingTasks && floatingTasks.length > 0 && (
-        <FloatingTaskDualView tasks={floatingTasks} />
+        <FloatingTaskDualView
+          tasks={floatingTasks}
+          onDeleteTask={handleDeleteTask}
+          deletingId={deletingId}
```
nxn-nicole: usetaskmutation has a isDeleting. if you use that one then don't need to use another deletingId

nxn-nicole (changes requested):

PR #49, opened 2025-11-26
Set up notification

nxn-nicole:


blotztask-mobile/src/app/(protected)/task-create.tsx
```diff
@@ -15,8 +17,12 @@ function TaskCreateScreen() {
   const handleTaskSubmit = async (formValues: TaskFormField) => {
     try {
       const dto = mapFormToAddTaskItemDTO(formValues);
-      await addTask(dto);
+      const newTaskId = await addTask(dto);
       posthog.capture("manual_task_creation");
+
+      const newTask = await fetchTaskById(newTaskId);
+      scheduleTaskReminder(newTask);
```
sol-wizard: why need another round trip to backend? i though one trip is enough ?

sol-wizard (changes requested):

blotztask-mobile/src/app/(protected)/task-create.tsx
nxn-nicole: changed

sol-wizard (approved):

PR #70, opened 2025-12-10
set up update user name

nxn-nicole:


blotztask-api/appsettings.json
```diff
@@ -33,8 +30,12 @@
     ]
   },
   "Auth0": {
-    "Domain": "dev-k72xachs0fr6nebp.us.auth0.com",
-    "Audience": "https://blotz-task-dev/api"
+    "Audience": "https://blotz-task-dev/api",
+    "Management": {
+      "Domain": "dev-k72xachs0fr6nebp.us.auth0.com",
```
sol-wizard: move this out from management

blotztask-api/Modules/Users/Commands/UpdateUserProfile.cs
```diff
@@ -0,0 +1,49 @@
...
+        db.AppUsers.Update(user);
+        await db.SaveChangesAsync(ct);
+
+        logger.LogInformation("User Profile for user {UserId} updated successfully in database.", profileCommand.Id);
+
+        await auth0ManagementService.UpdateUserProfileAsync(
+            auth0UserId,
+            profileCommand.DisplayName
+        );
+
```
sol-wizard, line 32: i think you should made changes to auth0 then call the sync handler

blotztask-api/Modules/Users/Commands/UpdateUserProfile.cs
```diff
@@ -0,0 +1,49 @@
...
+        logger.LogInformation("User Profile for user {UserId} updated successfully in auth0.", profileCommand.Id);
```
sol-wizard: but i think we dont want to call a handler in another handler (recursive) so find a better way?

blotztask-api/Modules/Users/Services/Auth0ManagementService.cs
```diff
@@ -0,0 +1,113 @@
...
+
+namespace BlotzTask.Modules.Users.Services;
+
+public interface IAuth0ManagementService
+{
+    Task UpdateUserProfileAsync(string auth0UserId, string? displayName,
+        CancellationToken ct = default);
+}
+
+public class Auth0ManagementService : IAuth0ManagementService
```
sol-wizard, line 19: try if auth0 sdk is better

sol-wizard (changes requested):

blotztask-api/Modules/Users/UserController.cs
```diff
@@ -36,14 +41,26 @@ public async Task<UserProfileDTO> GetUserProfile(CancellationToken ct)
...
-        return await getUserProfileQueryHandler.Handle(query, ct);;
-    }
-}
 
+        return await getUserProfileQueryHandler.Handle(query, ct);
+        ;
+    }
 
+    [HttpPut]
+    public async Task<string> UpdateProfile([FromBody] UpdateUserDto updateUserDto, CancellationToken ct)
```
sol-wizard: we usually do naming better, we should try to do controller and Dto and Command the same name

sol-wizard (approved):

PR #78, opened 2025-12-14
Feature/user preference

DuDuAu123:
### Summary
Add UserPreference entity with DB-level defaults and lazy-create support.

### Details
- Added UserPreference domain entity and EF configuration
- Enforced one row per user via PK on UserId
- Defaults enforced at database schema level
- Implemented lazy-create logic in UserPreferenceService to provision preferences on first access

### Verification
- Migration applied successfully
- Table created with correct defaults and constraints
image


blotztask-api/Modules/Users/Services/UserPreferenceService.cs
```diff
@@ -0,0 +1,44 @@
+using BlotzTask.Infrastructure.Data;
```
sol-wizard: i dont think we need the user preference service for now

blotztask-api/Modules/Users/DependencyInjection.cs
```diff
@@ -20,6 +20,7 @@ public static IServiceCollection AddUserModule(
         services.AddScoped<GetUserProfileQueryHandler>();
         services.AddScoped<UpdateUserProfileCommandHandler>();
         services.AddScoped<IAuth0ManagementService, Auth0ManagementService>();
+        services.AddScoped<UserPreferenceService>();
```
sol-wizard: i dont think we need user service for now

sol-wizard (changes requested):

blotztask-api/Modules/Users/DependencyInjection.cs
DuDuAu123: I've removed the UserPreferenceService and pushed the update. Thks

sol-wizard (approved):

PR #79, opened 2025-12-14
Set up settings screen based on design

nxn-nicole:


blotztask-mobile/src/app/(protected)/settings/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const handleProfileEdit = () => console.log("Edit profile pressed");
+
+  const handleMenuPress = (key: MenuKey) => {
+    if (key == "account") {
+      router.push("/(protected)/settings/account");
+    } else if (key == "task-handling") {
+      router.push("/(protected)/settings/task-handling");
+    } else if (key == "notifications") {
+      router.push("/(protected)/settings/notifications");
+    } else if (key == "under-development") {
```
sol-wizard: can we use something else like enum other than hard coded string ? and please use === instead of ==

sol-wizard: if this is more than one if statement, will switch case is better in this case?

blotztask-mobile/src/feature/settings/modals/menu-item.ts
```diff
@@ -0,0 +1,6 @@
+import { MaterialCommunityIcons } from "@expo/vector-icons";
+import { ComponentProps } from "react";
+
+type IconName = ComponentProps<typeof MaterialCommunityIcons>["name"];
+export type MenuKey = "account" | "task-handling" | "notifications" | "under-development";
```
sol-wizard: menu key is not a clear naming ? settingMenuKey?

sol-wizard (changes requested):

sol-wizard (approved):

PR #118, opened 2026-01-06
Add chinese language

Guotai812:


blotztask-mobile/src/i18n/index.ts
```diff
@@ -37,13 +39,15 @@ i18n.use(initReactI18next).init({
       settings: enSettings,
       tasks: enTasks,
       starSpark: enStarSpark,
+      "ai-task-generate": enAiTaskGenerate,
```
nxn-nicole: why not also use camelCase namespace

nxn-nicole (changes requested):

nxn-nicole (approved):

PR #142, opened 2026-01-14
starspark onboarding progress

LiChenyang-GZ:


sol-wizard:
this is not needed because we change onboarding method

PR #180, opened 2026-02-02
Feature/floating

Guotai812:


Guotai812:
https://github.com/user-attachments/assets/021c02c5-0ef3-4d31-91cb-831352e3b986

PR #193, opened 2026-02-10
refactor: update startTime, endTime and taskTimeType in task and deletedTask to required

Guotai812:


blotztask-api/Modules/Users/Commands/AddUser.cs
```diff
@@ -117,9 +117,9 @@ public async Task<AddUserResult> Handle(SyncUserCommand command, CancellationTok
                 {
                     Title = "Plan your first task",
                     Description = "This is a floating task example",
```
sol-wizard: change this description

sol-wizard (approved):

blotztask-api/Modules/Users/Commands/AddUser.cs
Guotai812: i delete it

sol-wizard (approved):

PR #220, opened 2026-02-23
ad log to management client

nxn-nicole:


sol-wizard (approved):

sol-wizard (commented):
Secret print ? No way

PR #229, opened 2026-03-05
Block addTask and editTask from persisting tasks with an empty/whites pace-only title in backend

LiChenyang-GZ:


blotztask-api/Modules/Tasks/Shared/TasTitleValidator.cs
```diff
@@ -0,0 +1,14 @@
+using System.ComponentModel.DataAnnotations;
+
```
nxn-nicole: you don't need to create a separate function for this. because this is just a small action. you can just deplicate the code in add task and edit task

nxn-nicole (changes requested):

blotztask-api/Modules/Tasks/Shared/TasTitleValidator.cs
LiChenyang-GZ: solved

PR #230, opened 2026-03-07
initial set up for reward related domain

Guotai812:


sol-wizard (commented):
lets call them as Badge not Rewards, in the folder

nxn-nicole (approved):

PR #236, opened 2026-03-10
feat: add TaskDeadline entity and establish a optional one-to-one rel…

HoshinoIzumi:
…ationship with TaskItem.


blotztask-api/Modules/Tasks/Domain/Entities/TaskDeadline.cs
```diff
@@ -0,0 +1,21 @@
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace BlotzTask.Modules.Tasks.Domain.Entities;
+
+public class TaskDeadline
+{
+    [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+    public int Id { get; set; }
+    
+    public int TaskItemId { get; set; }
```
sol-wizard, line 9: is this a required? you can add C# required https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/required

same for the rest of the field

blotztask-api/Modules/Tasks/Domain/Entities/TaskDeadline.cs
```diff
@@ -0,0 +1,21 @@
...
+    
+    public DateTimeOffset DueAt { get; set; }
+    
+    public DateTime CreatedAt { get; set; }
+    
+    public DateTime UpdatedAt { get; set; }
+    
+    public bool IsPinned { get; set; }
+    
+    public TaskItem TaskItem { get; set; } = null!;
```
sol-wizard: is this nullable ? do we even need it here ? double check it

sol-wizard (changes requested):

sol-wizard (approved):

PR #241, opened 2026-03-11
Input ai agent prompt and clean up

sol-wizard:


blotztask-api/Modules/Users/Commands/DeleteUser.cs
```diff
@@ -5,34 +5,39 @@
 
 namespace BlotzTask.Modules.Users.Commands;
 
+public class DeleteUserCommand
+{
+    public Guid UserId { get; set; }
+}
```
nxn-nicole, line 11: Why do we need to create a separate command class just to put in one property in it?

nxn-nicole (changes requested):

nxn-nicole (approved):

PR #243, opened 2026-03-12
feat: convert note to task - transactional endpoint and frontend integration

lyx2-2:
Exclude testing


blotztask-api/Modules/Notes/Commands/ConvertNoteToTask.cs
```diff
@@ -0,0 +1,90 @@
...
+        
+        // 1. search note using noteId + userId, not found -> 404
+        var note = await db.Notes.FirstOrDefaultAsync(
+            n => n.Id == command.NoteId && n.UserId == command.UserId,
+            ct);
+
+        if (note == null)
+            throw new NotFoundException("Note not found or no permission.");
+
+        var text = note.Text ?? string.Empty;
```
nxn-nicole: shouldn't allow an empty title write into task item table

blotztask-api/Modules/Notes/Commands/ConvertNoteToTask.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+        
+        // 3. create a new task Item
+        var newTask = new TaskItem
+        {
+            Title = title,
+            Description = description,
+            StartTime = command.StartTime,
+            EndTime = command.EndTime,
+            TimeType = TaskTimeType.RangeTime,
```
nxn-nicole: can not ensure it's range time. what if starttime = endtime?

blotztask-api/Modules/Notes/Controllers/NotesController.cs
```diff
@@ -82,4 +85,32 @@ public async Task<IActionResult> DeleteNote(Guid id, CancellationToken ct)
...
+            NoteId = id,
+            UserId = userId,
+            StartTime = request.StartTime,
+            EndTime = request.EndTime
+        };
+        
+        var newTaskId = await convertNoteToTaskCommandHandler.Handle(command, ct);
+        
+        var taskDto = await getTaskByIdQueryHandler.Handle(
+            new GetTasksByIdQuery { TaskId = newTaskId, UserId = userId },
```
nxn-nicole: seems like doesn't need the handler here? return task id is enough

blotztask-api/Modules/Notes/DTOs/ConvertNoteToTaskRequestDto.cs
```diff
@@ -0,0 +1,12 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace BlotzTask.Modules.Notes.DTOs;
+
+public class ConvertNoteToTaskRequestDto
```
nxn-nicole: for request dto, i think we usually just put it in  the command file?

blotztask-mobile/src/shared/hooks/add-note-to-task.ts
```diff
@@ -10,27 +13,46 @@ type AddNoteToTaskParams = {
   onSuccess?: () => void;
 };
 
+function invalidateSelectedDayTask(
```
nxn-nicole: i think we might have a similar function in blotztask-mobile/src/shared/hooks/useTaskMutations.ts, see whether you can reuse that

blotztask-mobile/src/shared/hooks/add-note-to-task.ts
```diff
@@ -10,27 +13,46 @@ type AddNoteToTaskParams = {
...
+  let cursorDay = startOfDay(start);
+  while (cursorDay <= end) {
+    queryClient.invalidateQueries({
+      queryKey: taskKeys.selectedDay(convertToDateTimeOffset(cursorDay)),
+    });
+    cursorDay = addDays(cursorDay, 1);
+  }
+}
+
 export const useAddNoteToTask = () => {
```
nxn-nicole: i think there is one more place you need to implement this function. check blotztask-mobile/src/feature/notes/components/note-time-estimate-modal.tsx, i have added a TODO there

blotztask-mobile/src/shared/hooks/add-note-to-task.ts
```diff
@@ -1,6 +1,9 @@
+import { useQueryClient } from "@tanstack/react-query";
```
nxn-nicole: i think the naming of this file has a little bit of issue. can check what other hook file name looks like and change like that

blotztask-mobile/src/shared/hooks/add-note-to-task.ts
```diff
@@ -10,27 +13,46 @@ type AddNoteToTaskParams = {
...
+    cursorDay = addDays(cursorDay, 1);
+  }
+}
+
 export const useAddNoteToTask = () => {
-  const { addTask } = useTaskMutations();
+  const queryClient = useQueryClient();
 
-  return ({ note, startTime, endTime, timeType = 1, onSuccess }: AddNoteToTaskParams) => {
+  return async ({ note, startTime, endTime, onSuccess }: AddNoteToTaskParams) => {
```
nxn-nicole: see how other react query do and try to handle loading state inside the hook and expose it to components

nxn-nicole (changes requested):

blotztask-mobile/src/shared/hooks/useAddNoteToTask.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export const useAddNoteToTask = () => {
+  const queryClient = useQueryClient();
+
+  const convertMutation = useMutation({
+    mutationFn: async ({ note, startTime, endTime }: AddNoteToTaskParams) => {
+      if (!note) throw new Error("No note");
+      const start = startTime ?? new Date();
+      const end = endTime ?? new Date(start.getTime() + 60 * 60 * 1000);
```
nxn-nicole: shouldn't add this error acceptance here. one day if some components pass in an empty value by mistake and instead of getting an error, you may find that it saved another value which could make bugs in app harder to find. shouldn't allow it to pass in null value

if this logic is for gashapon machine, you should put it in gashapon machine page

nxn-nicole (changes requested):

nxn-nicole (approved):

PR #244, opened 2026-03-12
Refactor the ai voice model

sol-wizard:


blotztask-mobile/src/feature/ai-task-generate/component/ai-tasks-preview.tsx
```diff
@@ -164,6 +159,9 @@ export function AiTasksPreview({
         )}
       </ScrollView>
 
+      {addError && (
```
nxn-nicole: idon't think we need to handle error ourselves, handle it with react query should be fine

nxn-nicole (changes requested):

nxn-nicole (approved):

PR #249, opened 2026-03-13
Add DDL Support to Task APIs and DDL Listing Endpoint

LiChenyang-GZ:


blotztask-api/Modules/Tasks/Commands/Tasks/AddTask.cs
```diff
@@ -45,6 +45,21 @@ public async Task<int> Handle(AddTaskCommand command, CancellationToken ct = def
         };
 
         db.TaskItems.Add(newTask);
+        
+        if (command.TaskDetails.IsDdl == true)
+        {
+            var deadline = new TaskDeadline
+            {
+                TaskItemId = newTask.Id,
```
HoshinoIzumi: Please telete this line. 
Setting `TaskItemId = newTask.Id` here assigns 0, since the entity hasn't been persisted yet. EF Core will resolve the foreign key through the TaskItem navigation property anyway, so this line has no real effect, but it's misleading to readers who might assume the ID is already valid.

blotztask-api/Modules/Tasks/Queries/Tasks/GetAllDdlTasks.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    public async Task<List<AllDdlTaskItemDto>> Handle(GetAllDdlTasksQuery query, CancellationToken ct = default)
+    {
+        logger.LogInformation("Fetching all DDL tasks for user {UserId}", query.UserId);
+
+        var ddlTasks = await db.TaskItems
+            .Include(t => t.Deadline)
+            .Include(t => t.Label)
+            .Where(t => t.UserId == query.UserId && t.Deadline != null)
+            .OrderByDescending(t => t.StartTime)
```
HoshinoIzumi: The current query sorts by `StartTime` descending, but for a DDL list users care about due date proximity, not start time. Suggest changing to `.OrderBy(t => t.Deadline!.DueAt)` so the nearest deadline appears first. Also, overdue tasks should be filtered out for current version (v1)

HoshinoIzumi (changes requested):
please check

blotztask-api/Modules/Tasks/Commands/Tasks/AddTask.cs
LiChenyang-GZ: deleted

blotztask-api/Modules/Tasks/Queries/Tasks/GetAllDdlTasks.cs
LiChenyang-GZ: OK

blotztask-api/Modules/Tasks/Commands/Tasks/AddTask.cs
```diff
@@ -45,12 +45,12 @@ public async Task<int> Handle(AddTaskCommand command, CancellationToken ct = def
         };
 
         db.TaskItems.Add(newTask);
+        await db.SaveChangesAsync(ct);
```
HoshinoIzumi: Please delete this line. We can use a single SaveChangesAsync instead. Add() just tracks changes in memory, and EF Core will automatically handle the insertion order via the navigation property `TaskItem = newTask`, resolving the foreign key before inserting the deadline. All changes are then wrapped in a single transaction in one call. The existing `SaveChangesAsync` on line 63 already handles this. Please refer to: https://learn.microsoft.com/en-us/ef/core/saving/basic#multiple-operations-in-a-single-savechanges

HoshinoIzumi (changes requested):

blotztask-api/Modules/Tasks/Commands/Tasks/AddTask.cs
LiChenyang-GZ: deleted

HoshinoIzumi (approved):

blotztask-api/Modules/Tasks/Commands/Tasks/EditTask.cs
```diff
@@ -47,6 +49,37 @@ public async Task<string> Handle(EditTaskCommand command, CancellationToken ct =
...
+                {
+                    task.Deadline = new TaskDeadline
+                    {
+                        TaskItemId = task.Id,
+                        TaskItem = task,
+                        DueAt = task.EndTime,
+                        CreatedAt = DateTime.UtcNow,
+                        UpdatedAt = DateTime.UtcNow,
+                        IsPinned = false
+                    };
```
nxn-nicole, line 65: find a way to simplify the ifs here. a lot of if looks strange

nxn-nicole (changes requested):

blotztask-api/Modules/Tasks/Controllers/TaskController.cs
```diff
@@ -109,6 +110,16 @@ public async Task<IEnumerable<AllTaskItemDto>> GetAllTasks(CancellationToken ct)
         var query = new GetAllTasksQuery { UserId = userId };
         return await getAllTasksQueryHandler.Handle(query, ct);
     }
+    
+    [HttpGet("ddl")]
```
nxn-nicole: I think you shouldn't put this DDL endpoint inside of TaskController. You can create a separate controller for deadline.

nxn-nicole (changes requested):

blotztask-api/Modules/Tasks/Queries/Deadlines/GetAllDdlTasks.cs
```diff
@@ -0,0 +1,78 @@
...
+            "Successfully fetched {TaskCount} DDL tasks for user {UserId}",
+            ddlTasks.Count,
+            query.UserId);
+
+        return ddlTasks;
+    }
+}
+
+public class AllDdlTaskItemDto
+{
```
nxn-nicole, line 60: This class naming sounds weird. Ask GPT to get a better name.

blotztask-api/Modules/Tasks/Queries/Tasks/GetAllDdlTasks.cs
```diff
@@ -0,0 +1,78 @@
...
+        
+        var now = DateTimeOffset.UtcNow;
+
+        var ddlTasks = await db.TaskItems
+            .Include(t => t.Deadline)
+            .Include(t => t.Label)
+            .Where(t => t.UserId == query.UserId 
+                        && t.Deadline != null
+                        && t.Deadline.DueAt >= now)
+            .OrderBy(t => t.Deadline!.DueAt)
```
nxn-nicole: I think the order by logic has a little bit of an issue here.

1. First, you need to return those deadlines where pinned is set to true.
2. Inside of that, it should be ordered by due date.
3. For other deadlines where pinned is false, they should also be ordered by due date.

blotztask-api/Modules/Tasks/Queries/Deadlines/GetAllDdlTasks.cs
```diff
@@ -0,0 +1,78 @@
...
+                StartTime = task.StartTime,
+                EndTime = task.EndTime,
+                IsDone = task.IsDone,
+                NotificationId = task.NotificationId,
+                AlertTime = task.AlertTime,
+                IsDdl = true,
+                DueAt = task.Deadline!.DueAt,
+                IsPinned = task.Deadline.IsPinned,
+                Label = task.Label != null
+                    ? new LabelDto
```
nxn-nicole, line 39: I think we only need to return those fields that we actually used in the frontend, which doesn't include:
1. Description
2. Alert time
3. Notification ID

blotztask-api/Modules/Tasks/Queries/Tasks/GetAllDdlTasks.cs
```diff
@@ -0,0 +1,78 @@
+using BlotzTask.Infrastructure.Data;
+using BlotzTask.Modules.Labels.DTOs;
+using Microsoft.EntityFrameworkCore;
+
+namespace BlotzTask.Modules.Tasks.Queries.Tasks;
```
nxn-nicole: Also for the place of this file, I think you can create a separate deadline folder inside of queries and then put all queries about deadlines there.

blotztask-api/Modules/Tasks/Queries/Tasks/GetAllDdlTasks.cs
```diff
@@ -0,0 +1,78 @@
...
+        logger.LogInformation("Fetching all DDL tasks for user {UserId}", query.UserId);
+        
+        var now = DateTimeOffset.UtcNow;
+
+        var ddlTasks = await db.TaskItems
+            .Include(t => t.Deadline)
+            .Include(t => t.Label)
+            .Where(t => t.UserId == query.UserId 
+                        && t.Deadline != null
+                        && t.Deadline.DueAt >= now)
```
nxn-nicole: I think instead of including those deadlines where the due date is later than now, you should only include those deadlines which haven't been finished.

blotztask-api/Modules/Tasks/Queries/Deadlines/GetAllDdlTasks.cs
```diff
@@ -0,0 +1,78 @@
...
+                    {
+                        LabelId = task.Label.LabelId,
+                        Name = task.Label.Name,
+                        Color = task.Label.Color
+                    }
+                    : null
+            })
+            .ToListAsync(ct);
```
nxn-nicole, line 48: You can add AsNoTracking to improve performance.

nxn-nicole (changes requested):

blotztask-api/Modules/Tasks/Commands/Tasks/EditTask.cs
LiChenyang-GZ: use switch

blotztask-api/Modules/Tasks/Controllers/TaskController.cs
LiChenyang-GZ: new a deadline controller

blotztask-api/Modules/Tasks/Queries/Deadlines/GetAllDdlTasks.cs
LiChenyang-GZ: change to DeadlineTaskDto

blotztask-api/Modules/Tasks/Queries/Tasks/GetAllDdlTasks.cs
LiChenyang-GZ: finished

blotztask-api/Modules/Tasks/Queries/Deadlines/GetAllDdlTasks.cs
LiChenyang-GZ: deleted

blotztask-api/Modules/Tasks/Queries/Tasks/GetAllDdlTasks.cs
LiChenyang-GZ: created

blotztask-api/Modules/Tasks/Queries/Tasks/GetAllDdlTasks.cs
LiChenyang-GZ: add where !t.IsDone

blotztask-api/Modules/Tasks/Queries/Deadlines/GetAllDdlTasks.cs
LiChenyang-GZ: added

blotztask-api/Modules/Tasks/Controllers/TaskController.cs
```diff
@@ -19,6 +20,7 @@ public class TaskController(
     DeleteTaskCommandHandler deleteTaskCommandHandler,
     EditTaskCommandHandler editTaskCommandHandler,
     GetAllTasksQueryHandler getAllTasksQueryHandler,
+    GetAllDdlTasksQueryHandler getAllDdlTasksQueryHandler,
```
nxn-nicole: do you still use this is task controller?

nxn-nicole (changes requested):

nxn-nicole (approved):

PR #250, opened 2026-03-13
fix chat history for ben

nxn-nicole:


blotztask-api/Modules/ChatTaskGenerator/Services/AiTaskGenerateService.cs
```diff
@@ -26,15 +26,16 @@ public async Task<AiGenerateMessage> GenerateAiResponse(ChatHistory chatHistory,
...
+
+            var result = await chatService.GetChatMessageContentAsync(
+                chatHistory,
+                executionSettings,
+                kernel,
+                ct
             );
 
+
             var responseContent = result.ToString();
```
sol-wizard, line 39: use toContent() ?

sol-wizard (approved):

PR #261, opened 2026-03-18
Feat/monthlt cal

Guotai812:


blotztask-api/Modules/Tasks/Queries/Tasks/GetMonthlyTaskAvailability.cs
```diff
@@ -0,0 +1,167 @@
...
+        
+        return result;
+    }
+    
+}
+
+public class MonthlyTaskIndicatorDto
+{
+    public DateTimeOffset Date { get; set; }
+    public bool HasTask { get; set; }
```
nxn-nicole: if you already return task list, is there need to return hastask as well? because this can be refered from taks list length

blotztask-api/Modules/Tasks/Queries/Tasks/GetMonthlyTaskAvailability.cs
```diff
@@ -0,0 +1,167 @@
...
+                    ))
+                .Select(t => new TaskThumbnailDto
+                {
+                    TaskTitle = t.Title,
+                    Label = t.Label
+                })
+                .ToList();
+            var hasTask = dayTasks.Count > 0;
+            
+            if (!hasTask)
```
nxn-nicole: I think we need to return at least three tasks for each day, so maybe it shouldn't be "hasTask" here. Maybe it's: when Task.length <3, go check recurring task.

nxn-nicole (changes requested):

nxn-nicole (approved):

PR #273, opened 2026-03-25
fix(mobile): resolve frontend bug bundle (#1209)

jen12138:
Done Fix Frontend Bugs (Language Update, Task Type Tab, Note Modal Behavior, Note Length Validation)


blotztask-mobile/src/feature/notes/screens/notes-screen.tsx
```diff
@@ -66,6 +68,11 @@ export default function NotesScreen() {
     const text = noteText.trim();
     if (!text) return;
 
+    if (text.length > 2000) {
+    setNoteError("Note text cannot exceed 2000 characters.");
+    return;
```
nxn-nicole: we don't handle error like this manually. we use react query to throw error and the error message will be automatically showed as blotztask-mobile/src/shared/components/ui/toast-config.tsx

blotztask-mobile/src/feature/notes/components/note-input-modal.tsx
```diff
@@ -45,6 +47,13 @@ export const NoteInputModal = ({
               autoFocus
               className="mt-3 h-36 rounded-xl bg-background px-4 text-base mx-4"
             />
+
+            {errorMessage ? (
+              <Text className="mt-2 mx-4 text-sm text-red-500 font-baloo">
+                {errorMessage}
+                </Text>
+            ) : null}
```
nxn-nicole: if you use toast config, you don't need to manually add error message ui

blotztask-mobile/src/feature/notes/components/note-input-modal.tsx
```diff
@@ -20,15 +21,16 @@ export const NoteInputModal = ({
...
   const isSaveEnabled = noteText.trim().length > 0;
   const { t } = useTranslation("notes");
 
   return (
     <Modal visible={visible} transparent animationType="fade" onRequestClose={onClose}>
       <KeyboardAvoidingView behavior="padding" className="flex-1">
-        <View className="flex-1 bg-black/30 justify-center items-center">
-          <View className="w-11/12 bg-white rounded-2xl p-3">
+        <Pressable className="flex-1 bg-black/30 justify-center items-center" onPress={onClose}>
+          <Pressable className="w-11/12 bg-white rounded-2xl p-3" onPress={() => {}}>
```
nxn-nicole: why need to add another empty function here

looks like we still can't close the note at the first background click

nxn-nicole (changes requested):

blotztask-mobile/src/feature/notes/components/note-input-modal.tsx
```diff
@@ -1,7 +1,7 @@
 import { theme } from "@/shared/constants/theme";
 import { MaterialCommunityIcons } from "@expo/vector-icons";
 import { useTranslation } from "react-i18next";
-import { ActivityIndicator, Modal, Pressable, Text, TextInput, View } from "react-native";
+import { ActivityIndicator, Modal, Pressable, Text, TextInput } from "react-native";
```
nxn-nicole: use react native modal - backdrop

nxn-nicole (changes requested):

jen12138:
Moved the note length validation out of `notes-screen` and into `notes-service` so it now throws through the existing React Query/toast flow. I also removed the manual error UI from `note-input-modal` and switched the modal to `react-native-modal` backdrop handling.

blotztask-mobile/src/feature/task-add-edit/components/segment-toggle.tsx
```diff
@@ -33,6 +33,14 @@ export function SegmentToggle({ value, setValue }: Props) {
   const tabAnimatedStyle = useAnimatedStyle(() => ({
     transform: [{ translateX: tabPositionX.value }],
   }));
+  useEffect(() => {
```
Guotai812: what does this effect do, i think it function the same as above. and i didnt see code fixing the bug in pbi, i think maybe the bug is not from this component

blotztask-mobile/src/feature/notes/services/notes-service.ts
```diff
@@ -24,17 +24,25 @@ export const deleteNote = async (noteId: string): Promise<void> => {
...
+    return await apiClient.post(url, { text: trimmedText });
   } catch {
     throw new Error("Failed to create note.");
   }
 };
 
 export const updateNote = async (editNoteDto: EditNoteDTO): Promise<NoteDTO> => {
   const url = `/notes/${editNoteDto.id}`;
+  const trimmedText = editNoteDto.text.trim();
+  if (trimmedText.length > 2000) {
```
Guotai812: we dont throw error here, backend will validate the length, we need to promt user when they enter the the content for note, you can think of when u have a overlengh name in other app, it promt user with a red color text, and doesnt allow user to input anymore

jen12138: Updated.
Moved the note length validation out of `notes-service.ts` and handled it in the input UI instead. The modal now shows an inline error message and prevents further input with `maxLength={2000}`. Save-time validation is kept as a fallback.

Guotai812 (approved):

blotztask-mobile/src/feature/notes/services/notes-service.ts
```diff
@@ -17,14 +17,24 @@ export const deleteNote = async (noteId: string): Promise<void> => {
 
 export const createNote = async (text: string): Promise<NoteDTO> => {
   const url = `/notes`;
-
-  return await apiClient.post(url, { text });
+  const trimmedText = text.trim();
+
```
nxn-nicole: I think the text has already been trimmed before passing into this service, so you don't need to trim it here again. 

Also, you don't need to add the try-catch block here because the return await will handle the try-catch error automatically.

nxn-nicole (changes requested):

jen12138 (commented):
Updated.

I removed the extra trim and try/catch in `notes-service.ts` as suggested. The note text is already handled before calling the service, and mutation errors are already surfaced through the existing React Query + toast flow.

nxn-nicole (approved):

PR #295, opened 2026-04-02
update pomodoro setting in backend

LiChenyang-GZ:


blotztask-api/Modules/Pomodoro/Controllers/PomodoroController.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.AspNetCore.Mvc;
+
+namespace BlotzTask.Modules.Users;
+
+[ApiController]
+[Route("api/[controller]")]
+[Authorize]
+public class PomodoroController(
+    UpdatePomodoroSettingCommandHandler updatePomodoroSettingCommandHandler,
+    ILogger<PomodoroController> logger) : Controller
```
nxn-nicole, line 12: can fix this warning

nxn-nicole (changes requested):

blotztask-api/Modules/Users/Commands/UpdatePomodoroSetting.cs
```diff
@@ -0,0 +1,47 @@
+using System.ComponentModel.DataAnnotations;
+using BlotzTask.Infrastructure.Data;
+using Microsoft.EntityFrameworkCore;
+
+namespace BlotzTask.Modules.Users.Commands;
```
nxn-nicole: i think you need to create a module for pomodoro setting and put everything related to pomodoro there. instead of leaving them inside user

nxn-nicole (changes requested):

blotztask-api/Modules/Users/Commands/UpdatePomodoroSetting.cs
LiChenyang-GZ: ok

blotztask-api/Modules/Pomodoro/Controllers/PomodoroController.cs
LiChenyang-GZ: OK

nxn-nicole (approved):

PR #308, opened 2026-04-11
add whisper

nxn-nicole:


blotztask-api/Modules/ChatTaskGenerator/Services/SpeechTranscriptionService.cs
```diff
@@ -1,77 +1,77 @@
...
+                    ResponseFormat = AudioTranscriptionFormat.Text
+                },
+                ct
+            );
 
-        using var response = await _http.SendAsync(request, ct);
-        var body = await response.Content.ReadAsStringAsync(ct);
 
-        if (!response.IsSuccessStatusCode)
+            var text = result.Value.Text;
```
sol-wizard: return result from whisper ai , please give better variable name

blotztask-api/Modules/ChatTaskGenerator/Services/SpeechTranscriptionService.cs
```diff
@@ -1,77 +1,77 @@
...
+
+
+            if (string.IsNullOrWhiteSpace(text))
+                throw new InvalidOperationException("Whisper transcription returned empty text.");
```
sol-wizard: this is not a good error message to return to frontend

sol-wizard (approved):

PR #328, opened 2026-04-23
fix setting page bug in ipad

AlexisEvan:
截屏2026-04-23 下午3 09 42

It has already been tested on ipad mini, after swiping up, the page has more white space, enough for users to click on the "about" button.


AlexisEvan:
截屏2026-04-23 下午3 14 12

Also checked in i Phone and was not significantly affected.

AlexisEvan:
Task : https://github.com/Blotz-Org/Blotz-Task-App-Private/issues/1283

nxn-nicole (approved):

PR #333, opened 2026-04-24
create add new subtask func

imlililili:
- when user click on "plus" button, it will create a new task with 0h0min duration

Simulator Screenshot - iPhone 17 Pro - 2026-04-24 at 20 00 27

### DIFF
--- a/blotztask-mobile/src/feature/task-details/components/subtasks-editor.tsx
+++ b/blotztask-mobile/src/feature/task-details/components/subtasks-editor.tsx
@@ -18,7 +18,7 @@ type SubtasksEditorProps = {
 
 const SubtasksEditor = ({ parentTask, onRefreshSubtasks, isRefreshingSubtasks }: SubtasksEditorProps) => {
   const { t } = useTranslation(["tasks", "common"]);
-  const { data: fetchedSubtasks, isLoading } = useSubtasksByParentId(parentTask.id);
+  const { data: fetchedSubtasks, isLoading } = useSubtasksByParentId(parentTask.id!);
 
   const { deleteSubtask, isDeletingSubtask, toggleSubtaskStatus, addSubtask, isAddingSubtask } = useSubtaskMutations();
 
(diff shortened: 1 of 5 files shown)


blotztask-mobile/src/feature/task-details/hooks/useSubtaskMutations.ts
```diff
@@ -105,6 +117,10 @@ export const useSubtaskMutations = () => {
     updateSubtask: updateSubtaskMutation.mutateAsync,
     isUpdatingSubtask: updateSubtaskMutation.isPending,
 
+    // Add subtask
+    addSubtask: addSubtaskMutation.mutateAsync,
```
nxn-nicole: Please use mutate instead of mutateAsync.

You can ask AI what's the difference between mutate and mutateAsync. I think mutate is better because it helps us handle the error at the same place.
Change the update subtask and other functions here to use mutate as well, if you need it.

blotztask-mobile/src/feature/task-details/components/subtasks-editor.tsx
```diff
@@ -49,7 +49,24 @@ const SubtasksEditor = ({ parentTask, onRefreshSubtasks, isRefreshingSubtasks }:
...
+    try {
+      await addSubtask({
+        parentTaskId: parentTask.id,
+        title: "New subtask",
+        duration: "00:00:00",
+        order: (fetchedSubtasks?.length ?? 0) + 1,
+      });
+    } catch (error) {
+      console.error("Failed to add subtask:", error);
+      Toast.show({
```
nxn-nicole: If you use mutate, there is no need to handle the error by yourself here.

blotztask-mobile/src/feature/task-details/hooks/useSubtaskMutations.ts
```diff
@@ -77,6 +78,17 @@ export const useSubtaskMutations = () => {
...
   });
 
+  const addSubtaskMutation = useMutation({
+    mutationFn: addSubtask,
+    onSuccess: (_, variables) => {
+      queryClient.invalidateQueries({ queryKey: subtaskKeys.all(variables.parentTaskId) });
+      queryClient.invalidateQueries({ queryKey: taskKeys.all });
+    },
+    onError: (error) => {
+      console.error("Failed to add subtask:", error);
```
nxn-nicole: If it's just the console log, you can remove this.

blotztask-mobile/src/feature/task-details/services/subtask-service.ts
```diff
@@ -16,6 +17,18 @@ export const createBreakDownSubtasks = async (
   }
 };
 
+export async function addSubtask(newSubtask: AddNewSubtaskDTO): Promise<number> {
+  const url = `${process.env.EXPO_PUBLIC_URL_WITH_API}/SubTask`;
```
nxn-nicole: I think now we have add ${process.env.EXPO_PUBLIC_URL_WITH_API} in the API client, so you don't need to add this in the service function again. Check if you still need this. if you don't need ${process.env.EXPO_PUBLIC_URL_WITH_API}, remove it from other functions in this file as well.

blotztask-mobile/src/feature/task-details/components/subtasks-editor.tsx
```diff
@@ -49,7 +49,24 @@ const SubtasksEditor = ({ parentTask, onRefreshSubtasks, isRefreshingSubtasks }:
...
+        type: "error",
+        text1: t("tasks:details.failedToAddSubtask"),
+      });
+    }
+  };
+
+  if (isLoading || isRefreshingSubtasks || isDeletingSubtask || isAddingSubtask) {
```
nxn-nicole: When I click "add a subtask," the whole subtask list flashes once. This isn't a good user experience. Check if you need this loading state in this way. For example, you can just disable the Add Subtask button when isAddingSubtask equals true.

nxn-nicole (changes requested):

nxn-nicole (approved):

PR #339, opened 2026-04-26
Clean up frontend

nxn-nicole:


blotztask-mobile/src/shared/models/task-detail-dto.ts
```diff
@@ -6,18 +6,21 @@ export enum TaskTimeType {
   Range = 1,
 }
 
-export interface TaskDetailDTO {
-  id: number | null;
-  recurringTaskId?: number | null;
-  description?: string;
+export interface TimedTaskFields {
```
sol-wizard: this is a weird naming what is timedtaskfields

looks like recommend from ai

sol-wizard (changes requested):

blotztask-mobile/src/shared/models/task-detail-dto.ts
nxn-nicole: i have changed

sol-wizard (approved):

PR #346, opened 2026-04-29
style: responsive UI zoom in

HoshinoIzumi:
allow scaling in 1.3x
pick one from either #346 or #347


HoshinoIzumi:
#347 is approved

PR #347, opened 2026-04-29
style: fixed UI zoom in

HoshinoIzumi:
make all font fixed
pick one choice from #346 and #347


blotztask-mobile/src/app/_layout.tsx
```diff
@@ -25,6 +25,22 @@ import Toast from "react-native-toast-message";
 import { toastConfig } from "@/shared/components/toast-config";
 import { useAuth } from "@/shared/hooks/useAuth";
 import posthog from "@/shared/constants/posthog-client";
+import { Text, TextInput } from "react-native";
+
+interface ComponentWithDefaultProps {
```
nxn-nicole: import { Text, TextInput } from "react-native";

type WithDefaultProps = { defaultProps?: { allowFontScaling?: boolean } };
(Text as unknown as WithDefaultProps).defaultProps = { allowFontScaling: false };
(TextInput as unknown as WithDefaultProps).defaultProps = { allowFontScaling: false };
Do you know the difference between type and interface in TypeScript?
I feel like using Type will be much easier than interface, so you don't need to extend all of the other realizations.

blotztask-mobile/src/app/_layout.tsx
```diff
@@ -25,6 +25,22 @@ import Toast from "react-native-toast-message";
...
+  defaultProps?: { allowFontScaling?: boolean };
```
nxn-nicole: Also, I feel like you can separate this into another file and just import it into the layout. 
I think we can disable all the system settings from the user's phone for now, such as whether the font is bold or the user's font preferences in their system settings. We should just use our own font, text size, and bold settings in our app.
This kind of responsiveness is not our app's focus, and most apps now do not allow users to change the font size from the outside.

nxn-nicole (changes requested):

blotztask-mobile/src/app/_layout.tsx
HoshinoIzumi: okay

nxn-nicole (approved):

PR #359, opened 2026-05-06
fixed Incorrect Default dueAt and is running normally locally

AlexisEvan:


AlexisEvan:
https://github.com/orgs/Blotz-Org/projects/1/views/2?pane=issue&itemId=180444844&issue=Blotz-Org%7CBlotz-Task-App-Private%7C1320

AlexisEvan:
The task has been resolved and is running normally locally.


截屏2026-05-06 下午9 35 26

blotztask-mobile/src/shared/models/base-task-dto.ts
```diff
@@ -12,4 +12,5 @@ export interface BaseTaskDTO {
   notificationId: string | null;
   alertTime?: string;
   isDeadline: boolean;
+  dueAt?: string;
```
nxn-nicole, line 15: i think we already have dueAt at TaskUpsertDTO if you have dueAt at BaseTaskDTO you can remove the dueat in TaskUpsertDTO

blotztask-api/Modules/Tasks/Queries/Tasks/GetTaskById.cs
```diff
@@ -31,6 +31,7 @@ public async Task<TaskByIdItemDto> Handle(GetTasksByIdQuery query, CancellationT
                 NotificationId = task.NotificationId,
                 AlertTime = task.AlertTime,
                 IsDeadline = db.TaskDeadlines.Any(td => td.TaskItemId == task.Id),
+                DueAt = task.Deadline == null ? null : task.Deadline.DueAt,
```
nxn-nicole: there are simple ways to write dueat and isdeadline
DueAt = task.Deadline?.DueAt, will make it simpler
there are also simpler ways to write 33 line isdeadline, because we already have left join in this query, there is simpler way to do this

nxn-nicole (changes requested):

blotztask-api/Modules/Tasks/Queries/Tasks/GetTaskById.cs
AlexisEvan: I tried DueAt = task.Deadline?.DueAt, but this projection is still an EF expression tree, so it fails to compile.

so I only simplified the 33 line

nxn-nicole (approved):

PR #361, opened 2026-05-10
numbers overflow

AlexisEvan:
截屏2026-05-10 下午4 12 02

The numbers are set within a designated area; increasing the number of digits slightly reduces the number's size to ensure it remains in the correct position.


AlexisEvan:
截屏2026-05-10 下午11 34 58

blotztask-mobile/src/feature/ddl/components/ddl-card.tsx
```diff
@@ -28,6 +27,17 @@ const DdlCard = ({ task }: { task: DeadlineTaskDTO }) => {
   } = useDdlMutation();
 
   const daysLeft = Math.max(0, differenceInCalendarDays(new Date(task.dueAt), new Date()));
+  const daysLeftText = daysLeft > 9999 ? "9999+" : daysLeft.toString();
+  const pinnedDaysFontSize =
```
nxn-nicole: adjusting the font size isn't good practice. you need to let the space allocate differently based on different length of the number instead of giving it different font size

nxn-nicole (changes requested):

AlexisEvan:
截屏2026-05-12 下午10 56 19

nxn-nicole (approved):

PR #385, opened 2026-05-25
Set up getting badge frontend

nxn-nicole:
Summary
Stored Expo push tokens per device — added a UserPushToken entity and POST /api/user-push-tokens endpoint that upserts a token by device ID, so each user's devices are tracked in the database.
Wired push token registration to the backend — on app launch the mobile app gets its Expo push token, diffs it against an AsyncStorage cache, and only calls the API when the token is new or changed.
Sent badge push notifications on task completion — added BadgeNotificationService which, after a task-completed event awards badges, looks up the user's stored tokens and sends a per-badge push via the Expo push API, respecting the user's language preference (EN/ZH).
Displayed badge notifications in-app — a new handleBadgeNotification handler shows an Alert with the badge name and description when a type: "badge" notification is received while the app is foregrounded.
Cleaned up the push notification hook — removed unused state (expoPushToken, channels, notification) from usePushNotificationSetup now that registration is a fire-and-forget side effect.
DB migration — added UserPushTokens table and an IsDisplayedToUserBadge column on UserBadge (for a future "catch-up on missed badge alerts" feature).
Notes / TODOs left in code
handleBadgeNotification: on Alert dismiss, call backend to mark badge as displayed (IsDisplayedToUserBadge = true).
On app launch, add a catch-up fetch for any IsDisplayedToUserBadge = false badges and re-surface them.

## Release note

> Earn badges as you complete tasks

**Status:**


blotztask-api/Modules/Users/Controllers/UserPushTokenController.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.AspNetCore.Mvc;
+
+namespace BlotzTask.Modules.UserPushTokens.Controllers;
+
+[Authorize]
+[ApiController]
+[Route("api/user-push-tokens")]
+public class UserPushTokenController(UpsertPushTokenCommandHandler upsertPushTokenCommandHandler) : ControllerBase
+{
+    public record UpsertPushTokenRequest(string Token, string DeviceId);
```
sol-wizard: Per project convention, `XRequest` classes belong in the command/query file, not the controller. Move `UpsertPushTokenRequest` next to `UpsertPushTokenCommand` in `Commands/UpsertPushToken.cs`.

blotztask-api/Modules/Tasks/Events/TaskCompletedEventHandler.cs
```diff
@@ -32,12 +36,15 @@ public async Task HandleAsync(TaskCompletedEvent taskCompletedEvent, Cancellatio
...
-        
-        await awardNewBadgesToUserHandler.Handle(new AwardNewBadgesToUserCommand
+
+        var awardedBadgeIds = await awardNewBadgesToUserHandler.Handle(new AwardNewBadgesToUserCommand
         {
             UserId = taskCompletedEvent.UserId,
             BadgeIds = matchingBadgeIds
         }, ct);
-      
+
```
sol-wizard: Debug `Console.WriteLine` left behind — remove, or replace with `logger.LogInformation` if you want it kept.

blotztask-api/Modules/Badges/Services/BadgeNotificationService.cs
```diff
@@ -0,0 +1,80 @@
...
+            body = isEnglish ? badge.NameEn : badge.NameZh,
+            data = new
+            {
+                type = "badge",
+                badgeId = badge.Id,
+                iconUrl = badge.IconUrl,
+                description = isEnglish ? badge.DescriptionEn : badge.DescriptionZh
+            }
+        });
+
```
sol-wizard, line 17: A 2xx response from Expo doesn't mean delivery succeeded — the response contains per-message tickets where individual entries can be `status: "error"` (e.g. `DeviceNotRegistered`). As written, stale/invalid tokens will silently fail forever and accumulate in the DB.

At minimum, log the response body on errors. Ideally parse tickets and delete tokens marked `DeviceNotRegistered`. OK to defer if tracked as follow-up.

blotztask-mobile/src/shared/services/notifications.ts
```diff
@@ -30,35 +16,56 @@ export async function registerForPushNotificationsAsync(): Promise<string | null
...
   }
+}
+
+async function handlePushTokenUpdate(token: string) {
+  const cachedToken = await AsyncStorage.getItem(STORAGE_KEYS.PUSH_TOKEN);
+  if (cachedToken === token) return;
+
+  const deviceId =
+    Platform.OS === "android"
+      ? Application.getAndroidId()
```
sol-wizard, line 66: Cache key is just the token, but Expo push tokens are device-scoped, not user-scoped. Scenario: user A logs out, user B logs in on the same device → token is unchanged → `cachedToken === token` short-circuits → backend row still has `UserId = A`, so user B will receive user A's badge notifications (and vice versa).

Fixes (pick one):
- Clear `STORAGE_KEYS.PUSH_TOKEN` on logout, or
- Key the cache as `${userId}:${token}`, or
- Just always upsert (the round-trip is cheap and it's the simplest fix).

blotztask-api/Modules/Users/Commands/UpsertPushToken.cs
```diff
@@ -0,0 +1,43 @@
...
+using BlotzTask.Infrastructure.Data;
+using BlotzTask.Modules.Users.Domain;
+using Microsoft.EntityFrameworkCore;
+
+namespace BlotzTask.Modules.UserPushTokens.Commands;
+
+public class UpsertPushTokenCommand
+{
+    [Required] public required string Token { get; init; }
+    [Required] public required string DeviceId { get; init; }
```
sol-wizard: nit: `[Required]` + `required` is redundant here. The command is constructed in the controller, not bound from the request, so the validation attribute doesn't fire anyway. If you want model validation, put `[Required]` on `UpsertPushTokenRequest` instead.

blotztask-mobile/src/shared/services/notifications.ts
```diff
@@ -30,35 +16,56 @@ export async function registerForPushNotificationsAsync(): Promise<string | null
...
-      finalStatus = status;
-    }
-    if (finalStatus !== "granted") {
-      alert("Failed to get push token for push notification!");
-      return null;
-    }
+  const settings = await Notifications.getPermissionsAsync();
+  let granted =
+    settings.granted || settings.ios?.status === Notifications.IosAuthorizationStatus.PROVISIONAL;
+
```
sol-wizard, line 23: nit: previous code had a `Device.isDevice` guard. Now on a simulator, `getExpoPushTokenAsync` will just throw and log an error. Probably fine, but worth keeping a clean early-return on simulator if you want to avoid noisy logs in dev.

sol-wizard (commented):
Full-stack PR (title says frontend, but includes the DB migration + push-token endpoint + Expo sender). Main things to address: request DTO location, leftover `Console.WriteLine`, no handling of Expo per-ticket errors, and the device-vs-user cache key in the mobile token sync. Details inline.

blotztask-api/Modules/Badges/DependencyInjection.cs
```diff
@@ -9,6 +10,13 @@ public static IServiceCollection AddBadgeModule(this IServiceCollection services
         services.AddScoped<FindMatchingBadgesHandler>();
         services.AddScoped<AwardNewBadgesToUserHandler>();
 
+        services.AddHttpClient("Expo", client =>
+        {
+            client.BaseAddress = new Uri("https://exp.host/--/api/v2/");
+            client.DefaultRequestHeaders.Add("Accept", "application/json");
+        });
+        services.AddScoped<IBadgeNotificationService, BadgeNotificationService>();
+
```
sol-wizard, line 21: Hardcoding the Expo base URL here isn't ideal:

- Can't swap it per environment (dev/staging/E2E mock).
- Expo recommends an access token for production push — that's a secret and must come from config, not code.
- Future tuning (timeouts, gzip, Polly retries) ends up scattered.

Suggest moving to `appsettings.json`:
```json
"Expo": {
  "BaseUrl": "https://exp.host/--/api/v2/",
  "AccessToken": ""
}
```
and binding via `IOptions<ExpoOptions>`.

Also prefer a **typed client** over a named one — the magic string `"Expo"` has to match exactly in two places and isn't discoverable:
```csharp
services.AddHttpClient<IExpoPushClient, ExpoPushClient>((sp, c) =>
{
    var opts = sp.GetRequiredService<IOptions<ExpoOptions>>().Value;
    c.BaseAddress = new Uri(opts.BaseUrl);
    c.DefaultRequestHeaders.Accept.Add(new("application/json"));
    if (!string.IsNullOrWhiteSpace(opts.AccessToken))
        c.DefaultRequestHeaders.Authorization = new("Bearer", opts.AccessToken);
});
```
Then `BadgeNotificationService` depends on `IExpoPushClient` instead of `IHttpClientFactory` — no string lookup, mockable in tests, Expo concern lives in one class.

blotztask-api/Modules/Badges/Services/BadgeNotificationService.cs
```diff
@@ -0,0 +1,80 @@
...
+public interface IBadgeNotificationService
+{
+    Task SendBadgeNotificationsAsync(Guid userId, List<int> badgeIds, CancellationToken ct = default);
+}
+
+public class BadgeNotificationService(
+    BlotzTaskDbContext db,
+    IHttpClientFactory httpClientFactory,
+    ILogger<BadgeNotificationService> logger) : IBadgeNotificationService
```
sol-wizard, line 12: This service is doing too much — it loads badges, tokens, *and* user preferences from `DbContext`, then builds the payload, then sends HTTP. That's three layers of responsibility in one class.

Suggested split (matches the layering pattern we already use for AI services — handler loads data, service takes primitives):

- The event handler (or a dedicated `AwardAndNotifyBadgesHandler`) does all the EF queries.
- This service becomes a thin transport: `SendAsync(tokens, title, body, data, ct)` — no `DbContext`, no language lookup.

Benefits: trivially unit-testable without an in-memory EF provider, and the Expo concern stops leaking into business logic.

blotztask-api/Modules/Badges/Services/BadgeNotificationService.cs
```diff
@@ -0,0 +1,80 @@
...
+            return;
+        }
+
+        var language = await db.UserPreferences
+            .Where(p => p.UserId == userId)
+            .Select(p => p.PreferredLanguage)
+            .FirstOrDefaultAsync(ct);
+
+        var isEnglish = language == Language.En;
+
```
sol-wizard: Localization (title string + `NameEn/NameZh` / `DescriptionEn/DescriptionZh` selection) shouldn't live as inline ternaries in a service. This is presentation logic leaking into infrastructure code.

Prefer `IStringLocalizer` / resx, and pass `language` in as a parameter rather than re-querying `UserPreferences` here.

blotztask-api/Modules/Badges/Services/BadgeNotificationService.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+        });
+
+        var client = httpClientFactory.CreateClient("Expo");
+        var response = await client.PostAsJsonAsync("push/send", messages, ct);
+
+        if (!response.IsSuccessStatusCode)
+        {
+            logger.LogError("Expo push failed with {Status} for user {UserId}, badges {BadgeIds}.",
+                response.StatusCode, userId, string.Join(", ", badgeIds));
```
sol-wizard: Only checking HTTP status isn't enough — Expo's `/push/send` returns a per-ticket array, and individual entries can be `status: "error"` with `details.error = "DeviceNotRegistered"` even when the HTTP response is 200.

Today those stale tokens stay in `UserPushTokens` forever and we'll keep pushing to dead devices. Recommend parsing the response and deleting tokens flagged `DeviceNotRegistered` / `InvalidCredentials`.

Also — this whole HTTP call is on the critical path of "user completes a task." A network hiccup with Expo and the user feels it. Should be queued (background channel / outbox) so task-complete returns immediately.

blotztask-api/Modules/Tasks/Events/TaskCompletedEventHandler.cs
```diff
@@ -32,12 +36,15 @@ public async Task HandleAsync(TaskCompletedEvent taskCompletedEvent, Cancellatio
...
+        Console.WriteLine("start sending new badges to user by notifications");
```
sol-wizard: `Console.WriteLine` — `logger` is already injected, use it (or drop the line; the next log inside the service already covers this).

blotztask-api/Modules/Tasks/Events/TaskCompletedEventHandler.cs
```diff
@@ -24,6 +26,8 @@ public async Task HandleAsync(TaskCompletedEvent taskCompletedEvent, Cancellatio
         if (task != null)
             completeOffsetMinutes = (int)(DateTimeOffset.UtcNow - task.EndTime).TotalMinutes;
         
+
+        // TODO: You can consider merging these three functions into one, so we won't need to use all three repetitively in every badge event handler.
```
sol-wizard: +1 on the TODO. Every future trigger handler will repeat this `find → award → notify` triplet. Worth collapsing into one `AwardAndNotifyBadgesHandler` now, before the second caller copies it.

sol-wizard (commented):
Review focused on service-layer business logic and the Expo HTTP client setup. Main asks: move Expo URL/token to config + use a typed client, slim down `BadgeNotificationService` so it only handles transport, and handle Expo's per-ticket errors so we can prune dead tokens.

blotztask-api/Modules/Badges/Services/BadgeNotificationService.cs
sol-wizard: Expanding on both points since they're separate issues bundled together:

### 1. "HTTP 200 doesn't mean the push worked"

Expo's `/push/send` returns 200 as long as it *received* the request. Per-device result is inside the JSON body:

```json
{
  "data": [
    { "status": "ok", "id": "xxx-ticket-id" },
    { "status": "error",
      "message": "\"ExponentPushToken[abc]\" is not a registered push token",
      "details": { "error": "DeviceNotRegistered" } }
  ]
}
```

So the HTTP call can be "successful" while every push silently failed.

**Why it matters:** when a user uninstalls the app, or the OS invalidates the token, Expo returns `DeviceNotRegistered` for that token forever. Without pruning:
- We waste network calls on every badge award to tokens that will never deliver.
- Expo's docs explicitly warn they'll rate-limit / block senders that keep pushing to dead tokens.

**Fix shape:** parse `response.data[]`, and for any entry with `details.error` in `("DeviceNotRegistered", "InvalidCredentials")`, delete the matching `UserPushToken` row.

### 2. "Don't make the user wait for Expo"

Current flow when a user completes a task:

```
user taps complete
  → TaskCompletedEventHandler
    → find badges (DB)
    → award badges (DB)
    → call Expo HTTP API   ← blocks here, 100ms–5s+
    → response to user
```

The user is waiting on the Expo round-trip just to see their task marked complete. Badge push isn't something the user is waiting for — it can arrive a couple seconds later, no one cares.

**Fix shape (simplest → most robust):**
- `Channel<T>` + a `BackgroundService` consumer (in-process, lost on crash — probably fine for badges).
- Outbox table — write a "send notification" row in the same DB transaction; a worker drains it (survives crashes, retryable).
- Hangfire / similar.

In-process channel is likely enough here.

sol-wizard (approved):

PR #392, opened 2026-05-27
fix build ps

nxn-nicole:


LiChenyang-GZ (approved):

blotztask-mobile/build.ps1
```diff
@@ -97,14 +97,24 @@ $easJson | ConvertTo-Json -Depth 10 | Set-Content $easPath
 # ------------------------
 # Build command
 # ------------------------
-$args = @(
+$buildArgs = @(
```
sol-wizard, line 100: Good catch renaming `$args` → `$buildArgs`. `$args` is a PowerShell automatic variable (the array of unbound arguments to a function/script), so the previous code could behave unpredictably depending on how the script is invoked. The rename is a real fix, not just a stylistic change.

blotztask-mobile/build.ps1
```diff
@@ -97,14 +97,24 @@ $easJson | ConvertTo-Json -Depth 10 | Set-Content $easPath
...
+# iOS: cloud build + auto-submit to App Store
+# Android APK: local build (sideload)
+# Android AAB: cloud build + auto-submit to Play Store
+if ($platform -eq "ios") {
+    $buildArgs += "--auto-submit"
+} elseif ($platform -eq "android" -and $androidBuildType -eq "apk") {
+    $buildArgs += "--local"
+} elseif ($platform -eq "android" -and $androidBuildType -eq "app-bundle") {
+    $buildArgs += "--auto-submit"
+}
```
sol-wizard, line 116: No `else` branch — if `$platform` is `android` but `$androidBuildType` is neither `apk` nor `app-bundle` (typo, empty, future value), the build silently runs as a cloud build with no `--auto-submit` and no `--local`. Probably worth either:

- Adding an `else { throw "Unknown android build type: $androidBuildType" }`, or
- Validating `$androidBuildType` earlier with a `ValidateSet` parameter attribute.

Same for `$platform` itself — anything other than `ios`/`android` falls through silently.

blotztask-mobile/build.ps1
```diff
@@ -97,14 +97,24 @@ $easJson | ConvertTo-Json -Depth 10 | Set-Content $easPath
...
+
+Write-Host "`nRunning: eas $($buildArgs -join ' ')" -ForegroundColor Green
 
-& eas @args
\ No newline at end of file
+& eas @buildArgs
```
sol-wizard, line 120: Nit: missing newline at end of file.

blotztask-mobile/build.ps1
```diff
@@ -97,14 +97,24 @@ $easJson | ConvertTo-Json -Depth 10 | Set-Content $easPath
...
+$buildArgs = @(
     "build"
     "--platform", $platform
     "--profile", $profile
     "--non-interactive"
-    "--local"
 )
 
-Write-Host "`nRunning: eas $($args -join ' ')" -ForegroundColor Green
+# iOS: cloud build + auto-submit to App Store
```
sol-wizard, line 107: The inline comment block here is helpful, but consider also calling out the behaviour change in the PR description: previously **all** builds ran with `--local`; now only Android APK does. iOS and Android AAB now run as cloud builds with auto-submit. Worth making explicit so reviewers/teammates know iOS builds no longer happen on the local machine.

sol-wizard (commented):
Good `$args` rename. Main ask: add an `else` to fail loudly on unknown `$platform`/`$androidBuildType` values instead of silently falling through.

PR #395, opened 2026-05-28
refactor: update default query options for improved cache freshness

HoshinoIzumi:
Remove `refetchOnMount: "always"`: This was overriding `staleTime` and forcing a network call on every screen mount. Removing it lets the 30s freshness window take effect — rapid screen navigation (< 30s) now skips refetch.
`gcTime: 1min` → `5min`: Prevents cache from being garbage-collected during brief app switches.
Smart retry: 4xx errors no longer retry, since they are deterministic. 5xx and network errors still retry up to 2 times with exponential backoff.

The main UX impact: Cache is retained for 5 minutes instead of 1 minute after leaving a screen, so brief app switches no longer cause a cold reload. Client errors (4xx) now fail immediately instead of retrying twice. Note: in React Native, returning to a previous screen via stack pop does not trigger remount, so refetchOnMount was never involved in that flow — behavior there is unchanged.


blotztask-mobile/src/shared/util/queryClient.ts
```diff
@@ -48,11 +48,19 @@ export const queryClient = new QueryClient({
   }),
   defaultOptions: {
     queries: {
-      staleTime: Infinity,
-      gcTime: 1 * 60 * 1000,
-      retry: 2,
+      staleTime: 30_000,
```
sol-wizard: Worth being explicit in the PR description: previously `staleTime: Infinity` meant data was only ever refetched on explicit `invalidateQueries`. With `30_000`, anything older than 30s will refetch on mount / focus / reconnect.

The PR title says "improved performance," but this trades fewer requests for fresher data — net request volume will likely **increase**, not decrease. The actual perf win comes from removing `refetchOnMount: "always"` (combined with the 30s freshness window, rapid screen navigation now skips refetch). Might be worth retitling to "improve cache freshness" or similar to set expectations.

blotztask-mobile/src/shared/util/queryClient.ts
```diff
@@ -48,11 +48,19 @@ export const queryClient = new QueryClient({
...
+      retry: (failureCount, error) => {
+        if (
+          isAxiosError(error) &&
+          error.response?.status &&
+          error.response.status < 500
+        ) {
+          return false;
+        }
+        return failureCount < 2;
+      },
```
sol-wizard, line 59: Good change — skipping retries on 4xx avoids hammering the backend on auth or validation errors.

Two thoughts:
1. **Consider exempting 408 (Request Timeout) and 429 (Too Many Requests)** — both are 4xx but typically *are* worth retrying (often with backoff). Right now they're treated as terminal.
2. The status check is a bit verbose. Since `error.response?.status` is `number | undefined`, you can simplify:
   ```ts
   const status = isAxiosError(error) ? error.response?.status : undefined;
   if (status && status < 500 && status !== 408 && status !== 429) return false;
   return failureCount < 2;
   ```
   Also handles the network-error case (no `response`) cleanly — those will retry, which is probably what you want.

blotztask-mobile/src/shared/util/queryClient.ts
```diff
@@ -48,11 +48,19 @@ export const queryClient = new QueryClient({
...
       retryDelay: (attempt) => Math.min(1000 * 2 ** attempt, 8000),
-      refetchOnMount: "always",
       refetchOnReconnect: true,
```
sol-wizard, line 61: Removing `refetchOnMount: "always"` is the most consequential change here for UX — any screen that previously assumed "always-fresh on mount" (e.g. pull-to-refresh that just remounts, post-mutation navigation that expects new data) will now show cached data for up to 30s.

Sanity check: are there any flows (notification badges, task lists right after create/update) that rely on the old always-refetch behavior? If yes, those need explicit `invalidateQueries` calls or per-query `staleTime: 0`.

sol-wizard (commented):
Solid direction. Main asks: clarify PR title (this is about freshness, not raw performance), exempt 408/429 from the no-retry path, and sanity-check any flows that depended on the old always-refetch-on-mount behavior.

blotztask-mobile/src/shared/util/queryClient.ts
HoshinoIzumi: no flows rely on the old always-refetch behavior.

- Notification badges: App doesn't use badge counts (`shouldSetBadge: false`).
- Pull-to-refresh: No `RefreshControl` anywhere in the codebase.
- Post-mutation navigation: All mutation hooks call `invalidateQueries` in `onSuccess`, which ignores `staleTime` and forces a refetch:
  -  `useTaskMutations` → invalidates taskKeys.all + selectedDay + ddl
  -  `useSubtaskMutations` → invalidates subtaskKeys + taskKeys
  -  `useNotesMutation` → invalidates noteKeys.all
  -  `useDdlMutation` → invalidates ddlKeys + taskKeys
  -  `useUserProfileMutation / useUserPreferencesMutation` → invalidates respective keys
  -  `useFocusEffect`: Only used in notes-screen.tsx for analytics tracking, not data fetching.

- Optimistic updates: `task-card.tsx` pre-fills cache via setQueryData before navigating to details, also not dependent on mount refetch.

blotztask-mobile/src/shared/util/queryClient.ts
HoshinoIzumi: now we move it back to Infinity, because as long as user do not update task/data, the task/data should be accurate

blotztask-mobile/src/shared/util/queryClient.ts
HoshinoIzumi: We won't get 408 — our backend doesn't return it, and Axios timeouts produce a network error (no response), not a 408 status code. 
As for 429, it's only triggered by AiQuotaExceededException when the user's monthly AI token quota is exceeded. This is a monthly limit, so retrying after a few seconds won't help. If we add short-window rate limiting in the future, we can revisit this.

sol-wizard (approved):

PR #397, opened 2026-05-28
 add force update support

Moomin928:


blotztask-api/Modules/AppVersion/DependencyInjection.cs
```diff
@@ -0,0 +1,17 @@
...
+
+
+namespace BlotzTask.Modules.AppVersion
+{
+    public static class DependencyInjection
+    {
+        public static IServiceCollection AddAppVersionModule(this IServiceCollection services)
+        {
+            services.AddOptions<AppVersionOptions>()
+                .BindConfiguration("AppVersion");
```
sol-wizard: 🔴 **Blocker:** `AddAppVersionModule()` is never called in `Program.cs`. Other modules are registered at lines 32–41 — this one is missing, so the options binding never runs and the endpoint will ship returning empty strings for every field.

blotztask-api/Modules/AppVersion/Options/AppVersionOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public PlatformVersionPolicy Ios {get;set;} = new();
+
+        public PlatformVersionPolicy Android {get;set;} = new();
+    }
+
+    public sealed class PlatformVersionPolicy
+    {
+        public string LatestVersion {get;set;} = "";
+
```
sol-wizard, line 13: 🟡 **Config source — recommendation, not a strict blocker.**

The DI registration above is still a blocker, but where the values *live* is a separate decision. My recommendation:

**Staging / Production → App Service environment variables (Configuration → Application settings)**, not `appsettings.json`. Reason: bumping `MinimumSupportedVersion` after a forced rollout shouldn't require a code change + PR + redeploy. With env vars you just update the value in the Azure portal and restart the app.

ASP.NET's config system reads env vars using `__` as the section separator, so the existing `BindConfiguration("AppVersion")` picks them up with zero code change. Set them as:

```
AppVersion__Ios__LatestVersion             = 1.2.0
AppVersion__Ios__MinimumSupportedVersion   = 1.0.0
AppVersion__Ios__StoreUrl                  = https://apps.apple.com/...
AppVersion__Android__LatestVersion         = 1.2.0
AppVersion__Android__MinimumSupportedVersion = 1.0.0
AppVersion__Android__StoreUrl              = https://play.google.com/...
```

**Local development → `appsettings.Development.json`.** This file is loaded only when `ASPNETCORE_ENVIRONMENT=Development`, so it won't leak into staging/prod. Devs get sensible defaults without having to set env vars on their machine.

**Why this works — config provider priority:** ASP.NET loads sources in order, and later sources override earlier ones. The default order is `appsettings.json` → `appsettings.{Environment}.json` → user secrets (Dev only) → **environment variables** → command-line args. So env vars always win over the JSON files, which is exactly what we want for the prod gate.

Docs:
- [App Service app settings (env vars)](https://learn.microsoft.com/en-us/azure/app-service/configure-common?tabs=portal#configure-app-settings) — how to set them in the portal and the `__` nesting convention
- [Configuration providers & precedence in ASP.NET Core](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-9.0#configuration-providers) — the load order above
- [Use multiple environments (`appsettings.{Environment}.json`)](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-9.0) — how Development-only config works

Concrete ask before merge:
1. Add a dev-only `AppVersion` section to `appsettings.Development.json` with placeholder values.
2. Set the real values as App Service application settings in staging + prod (no code change needed).
3. Keep `appsettings.json` clean — no `AppVersion` section there, so it's obvious the values come from env.

Also still applies from before: typo `MinimumSupportVersion` → `MinimumSupportedVersion`, drop the unused `using` block, add trailing newline.

blotztask-api/Modules/AppVersion/Controller/AppVersionController.cs
```diff
@@ -0,0 +1,30 @@
...
+    private readonly AppVersionOptions _options;
+
+    public AppVersionController(IOptions<AppVersionOptions> options)
+    {
+        _options = options.Value;
+    }
+    
+    [HttpGet]
+    public ActionResult<AppVersionOptions> Get()
+    {
```
sol-wizard, line 16: Returning the raw `AppVersionOptions` class as the response shape couples the wire contract to the internal config class — renaming a property later becomes a breaking client change. Consider a thin response DTO. (Camel-casing is already handled by `System.Text.Json` defaults, so that part's fine.)

blotztask-mobile/src/shared/services/version-service.ts
```diff
@@ -3,41 +3,47 @@ import { Platform } from "react-native";
...
+  ios: PlatformVersionPolicy;
+  android: PlatformVersionPolicy;
+};
+
 export const checkForUpdate = async (): Promise<UpdateCheckResult> => {
   const isProduction = process.env.EXPO_PUBLIC_APP_ENV === "production";
-  const isIOS = Platform.OS === "ios";
+  if (!isProduction) return { isUpToDate: true, isForceUpdate: false, storeUrl: null };
 
   // TODO: Skip andoid check for now, need to figure out the best way to do it
```
sol-wizard: 🟡 Android is now active in production — previously it was explicitly skipped (`!isIOS`). The `// TODO: Skip andoid check for now` comment above is now stale and misleading. Confirm the Android `storeUrl` / `minimumSupportVersion` are actually populated in config before this ships, otherwise Android users get force-updated against empty strings. (ask Nicole to confirm)

blotztask-mobile/src/shared/services/version-service.ts
```diff
@@ -3,41 +3,47 @@ import { Platform } from "react-native";
 
 export type UpdateCheckResult = {
   isUpToDate: boolean;
+  isForceUpdate: boolean;
```
sol-wizard: 🟡 **Suggestion: collapse the two booleans into a single \`status\` field.**

Right now the function returns two flags:

```ts
type UpdateCheckResult = {
  isUpToDate: boolean;
  isForceUpdate: boolean;
  storeUrl: string | null;
};
```

With two independent booleans there are **four possible combinations**, but only three are real states:

| `isUpToDate` | `isForceUpdate` | meaning            |
| ------------ | --------------- | ------------------ |
| `true`       | `false`         | up to date ✅      |
| `false`      | `false`         | soft update ✅     |
| `false`      | `true`          | force update ✅    |
| `true`       | `true`          | ❓ impossible — but the type allows it |

That last row is a bug waiting to happen. If someone later writes `result.isUpToDate && !result.isForceUpdate`, the compiler can't help them because the type says it's legal.

**Fix:** return one tagged field instead of two booleans, the same way `UpdateCheckState` is already modelled in `useUpdateCheck.ts`:

```ts
export type UpdateCheckResult =
  | { status: "upToDate" }
  | { status: "outdated";    storeUrl: string }
  | { status: "forceUpdate"; storeUrl: string };
```

Now `checkForUpdate` returns one of exactly three shapes — the impossible fourth state can't be constructed. And the hook becomes a near-passthrough:

```ts
// useUpdateCheck.ts — before
.then(({ isUpToDate, isForceUpdate, storeUrl }) => {
  if (isForceUpdate) {
    setState({ status: UpdateCheckStatus.ForceUpdate, storeUrl: storeUrl ?? "" });
  } else if (isUpToDate) {
    setState({ status: UpdateCheckStatus.UpToDate });
  } else {
    setState({ status: UpdateCheckStatus.Outdated, storeUrl: storeUrl ?? "" });
  }
});

// after
.then(setState);
```

The service already knows which of the three states it's in — having it return that directly removes the if/else translation layer in the hook and makes the two layers use the same vocabulary.

This pattern is called a **discriminated union** (the `status` field is the "discriminant" that tells TypeScript which variant you're in). It's the same idea as `UpdateCheckState` you already wrote — this comment is just suggesting we use it in `version-service.ts` too, instead of two booleans.

---

**Side note: drop the \`UpdateCheckStatus\` enum, use string literals directly.**

You may be wondering: \`"upToDate"\` is a hardcoded string — shouldn't we use the existing \`UpdateCheckStatus\` enum instead? My recommendation is **no, go the other way**: replace the enum with string literals everywhere.

Reasons:

1. **Enums emit runtime JavaScript.** \`enum UpdateCheckStatus { ... }\` compiles into a real object that ships in the bundle. A string union is purely a compile-time type — zero runtime cost, zero bytes shipped.
2. **String unions are what-you-see-is-what-you-get.** \`status: "upToDate"\` is the value in code, at runtime, and in any log line. With an enum there's an extra indirection (\`UpdateCheckStatus.UpToDate\` → \`"upToDate"\`) you have to remember.
3. **The TypeScript team itself recommends string literal unions over enums** for most cases. \`const enum\` has known issues with \`isolatedModules\` (which Metro/Expo use), and regular enums have some surprising assignability rules.
4. **Narrowing works identically.** \`if (result.status === "forceUpdate")\` gives the same exhaustiveness checks as the enum version — TypeScript narrows the union just as well.

So the full proposal is:

```ts
// version-service.ts
export type UpdateCheckResult =
  | { status: "upToDate" }
  | { status: "outdated";    storeUrl: string }
  | { status: "forceUpdate"; storeUrl: string };

// useUpdateCheck.ts — delete the enum, reuse the same type
export type UpdateCheckState =
  | { status: "pending" }
  | UpdateCheckResult;
```

The important thing is **don't mix** — string union in the service + enum in the hook would be the worst of both worlds. Pick one vocabulary across both layers.

blotztask-mobile/src/shared/components/update-required-screen.tsx
```diff
@@ -3,14 +3,19 @@ import { Image } from "expo-image";
...
   storeUrl: string;
+  forced: boolean;
 };
 
-export function UpdateRequiredScreen({ storeUrl }: Props) {
+export function UpdateRequiredScreen({ storeUrl, forced }: Props) {
   const { t } = useTranslation("common");
-
+  const route = useRouter();
+  const skip = () => {
```
sol-wizard, line 18: - `route` → `router` (it's the router instance, not a route).
- Hardcoded `replace("/(auth)/signin")` assumes the user is unauthenticated when they reach this screen. That's true today (only `index.tsx` redirects here), but it's an implicit coupling — a `router.back()` would be safer, or branch on auth state.
- The title/message still read as a soft "Fresh Update Dropped!" even when `forced`. Consider `update.forced.title` / `update.forced.message` for the hard-block case.

blotztask-api/Modules/AppVersion/DependencyInjection.cs
```diff
@@ -0,0 +1,17 @@
...
+            return services;
+        }
+        
+
+    }
```
sol-wizard, line 12: Nit: this file uses old-style braced namespace while the controller uses file-scoped. Match the rest of the module. Also missing trailing newline.

sol-wizard (changes requested):
Force-update flow is well-structured, but two blockers must be fixed before merge:

1. `AddAppVersionModule()` is never registered in `Program.cs`.
2. No `AppVersion` section exists in `appsettings*.json`.

Together these mean the endpoint ships returning empty strings and the client's version comparison is undefined. See inline comments for details and smaller suggestions.

Moomin928:
Simulator Screenshot - iPhone 17 Pro Max - 2026-05-28 at 21 50 58

Moomin928:
Simulator Screenshot - iPhone 17 Pro Max - 2026-05-28 at 22 39 48

blotztask-api/Modules/AppVersion/Controller/AppVersionController.cs
Moomin928: done

blotztask-api/Modules/AppVersion/Options/AppVersionOptions.cs
Moomin928: done

blotztask-api/Modules/AppVersion/DependencyInjection.cs
Moomin928: done

blotztask-mobile/src/shared/components/update-required-screen.tsx
Moomin928: during to the index use redirect ，the router.back() can not work well,so i change to auth state

blotztask-api/Modules/AppVersion/Controller/AppVersionController.cs
```diff
@@ -0,0 +1,26 @@
...
+    public ActionResult<AppVersionResponse> Get()
+    {
+        var option = options.Value;
+        return Ok(new AppVersionResponse(
+            Ios: Map(option.Ios),
+            Android: Map(option.Android)
+        ));
+    }
+
+    private static PlatformVersionPolicyResponse Map(PlatformVersionPolicy p) =>
```
sol-wizard, line 24: is this just mapping ? if so i will perfer do it in the controller , not need for split into individual function. easier to read

blotztask-api/Modules/AppVersion/Options/AppVersionOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+    public sealed class AppVersionOptions
+    {
+        public PlatformVersionPolicy Ios {get;set;} = new();
+
+        public PlatformVersionPolicy Android {get;set;} = new();
+    }
+
+    public sealed class PlatformVersionPolicy
+    {
+        public string LatestVersion {get;set;} = "";
```
sol-wizard, line 12: i feel like we should make this field required, because it will cause problem if any reason this is null (same as field below)

sol-wizard (changes requested):

sol-wizard (approved):

PR #419, opened 2026-06-11
Generalize Monthly Review into a period-based Review feature (adds Weekly)

sol-wizard:
## Summary

Generalizes the shipped Monthly Review into a single Review feature that supports both Weekly and Monthly periods on the backend.

## Release note

Generalizes the shipped Monthly Review into a single Review feature that supports both Weekly and Monthly periods on the backend. User can now have weekly review in the app
Your reviews just got more frequent — Blotz now writes you a weekly letter reflecting on your week, alongside your monthly one.

**Status:**


blotztask-api/Modules/Reviews/Commands/GenerateReview.cs
```diff
@@ -246,10 +251,10 @@ private Task<Language> LoadPreferredLanguageAsync(Guid userId, CancellationToken
...
         {
-            logger.LogWarning(ex, "Unexpected error while generating monthly review");
+            logger.LogWarning(ex, "Unexpected error while generating review");
             throw new AiTaskGenerationException(
                 AiErrorCode.Unknown,
-                "An unhandled exception occurred during monthly review generation.",
+                "An unhandled exception occurred during review generation.",
                 ex);
         }
     }
```
Xu-create-ops, line 260: Please consider concurrency here. The code first checks whether a report exists; if not, it executes the “Call AI Generation” operation, and finally writes to the database. This works fine in a single-threaded environment, but problems arise when two requests arrive simultaneously.

Guotai812:
The MonthlyReviewScreen has both weekly and monthly, maybe change it to ReviewScreen

Xu-create-ops (approved):

nxn-nicole (approved):

blotztask-api/Modules/Reviews/Commands/GenerateReview.cs
sol-wizard: will resolve that in the next branch

PR #420, opened 2026-06-11
Recurring task

Xu-create-ops:
## Release note

Status:
- [ ] User-facing
- [x] Beta
- [ ] Hidden in production
- [ ] Internal only

Summary:
Add recurring task support across backend APIs, calendar/deadline queries, and mobile task flows.

Details:
This PR introduces the upgraded recurring task model with recurring task series, per-occurrence overrides, virtual occurrence materialization, and support for editing or deleting either a single occurrence or future occurrences.

It also updates calendar, deadline, weekly, and monthly task queries so recurring virtual, materialized, modified, skipped, completed, and detached occurrences are represented consistently.

On mobile, this adds recurring task creation payloads, mutation hooks, recurring task identity helpers, virtual task detail navigation, and action sheets for choosing whether edits or deletes apply to one occurrence or future occurrences.


blotztask-mobile/src/feature/task-add-edit/screens/task-edit-screen.tsx
```diff
@@ -49,8 +334,104 @@ export default function TaskEditScreen() {
...
+
+  return {
+    frequency: frequencyMap[recurrence],
+    interval: recurrence === "biweekly" ? 2 : 1,
+    daysOfWeek:
+      recurrence === "weekly" || recurrence === "biweekly"
+        ? getWeeklyDayFlag(occurrenceDate)
+        : null,
+    dayOfMonth: recurrence === "monthly" ? Number(occurrenceDate.slice(8, 10)) : null,
+  };
```
sol-wizard, line 447: **Major — silent loss of `interval` and `daysOfWeek` on edit-future.**

The UI collapses the backend's `RecurrencePattern` into 5 string buckets (`daily/weekly/biweekly/monthly/yearly`) and rebuilds the pattern from scratch on submit:

- Any template with `interval ∉ {1, 2 for Weekly}` round-trips as `interval=1`. A "every 3 weeks" or "every 6 months" template (supported by the backend, the API DTO, and tests) gets silently flattened to "every week" / "every month" the first time the user hits "edit future".
- `daysOfWeek` is rebuilt from the occurrence date alone via `getWeeklyDayFlag(occurrenceDate)`, which destroys multi-day weekly patterns (e.g. Mon+Wed+Fri → just whatever day the edited occurrence falls on).

The UI can't *create* these patterns today, so production data is safe right now, but `RecurringTaskEditMetadataDto` already carries the real `Interval` / `DaysOfWeek` to the client. As soon as a "custom" recurrence picker (or any non-mobile creator) lands, every subsequent edit will corrupt the rule.

Suggest either:
1. Round-tripping `interval` / `daysOfWeek` / `dayOfMonth` from `selectedTask.recurringTask` when the user hasn't changed the recurrence dropdown, or
2. Blocking the "edit future" path entirely when the existing pattern isn't representable in the dropdown and surfacing a "this recurrence rule isn't editable in the app yet" message.

blotztask-api/Modules/Tasks/Queries/Tasks/GetTasksByDate.cs
```diff
@@ -84,7 +84,15 @@ public async Task<List<TaskByDateItemDto>> Handle(GetTasksByDateQuery query, Can
...
-                RecurringTaskId = task.RecurringTaskId,
+                OccurrenceKind = task.RecurringOccurrenceOverride == null
+                    ? TaskOccurrenceKind.NormalTaskItem
+                    : TaskOccurrenceKind.MaterializedRecurringOccurrence,
+                RecurringOccurrence = task.RecurringOccurrenceOverride != null
+                    ? TaskOccurrenceDtoHelpers.ToRecurringOccurrenceIdentity(task.RecurringOccurrenceOverride)
+                    : null,
+                RecurringTask = task.RecurringOccurrenceOverride != null
+                    ? TaskOccurrenceDtoHelpers.ToRecurringTaskMetadata(task.RecurringOccurrenceOverride.RecurringTask)
+                    : null,
```
sol-wizard: **Major — `Detached` occurrences are reported inconsistently across queries.**

Here, any `TaskItem` with an attached override (including `Detached`) is projected as `OccurrenceKind = MaterializedRecurringOccurrence` with `RecurringOccurrence = ToRecurringOccurrenceIdentity(override)` populated.

`GetAllDdlTasks` does the opposite — it hard-codes `NormalTaskItem` with `RecurringOccurrence = null` for the same row (see `GetAllDdlTasks.cs:31-33`).

The mobile side uses `getRecurringOccurrenceIdentity(task)` (`task-card.tsx:155`, `task-edit-screen.tsx:131`) to decide whether to show the "this occurrence / future occurrences" action sheet. So a detached task will:
- on the calendar → route through the recurring delete/edit sheet and ask the user about "future occurrences" of a series they already stopped, then call the recurring endpoints (which will then fail because the template is truncated past the date), and
- on the DDL page → behave correctly as a normal task.

Pick one — I think the DDL behavior is the right one (detached *is* a normal task, by definition) and this projection should fall back to `NormalTaskItem` + `RecurringOccurrence = null` when `OverrideType == Detached`.

blotztask-api/Modules/Tasks/Domain/Services/RecurringOccurrenceMaterializer.cs
```diff
@@ -0,0 +1,146 @@
...
+            UpdatedAt = DateTime.UtcNow
+        };
+
+        recurringOverride.TaskItem = taskItem;
+        recurringOverride.RecurringTaskId = template.Id;
+        recurringOverride.OverrideType = RecurringOccurrenceOverrideType.Materialized;
+        recurringOverride.UpdatedAt = DateTime.UtcNow;
+        taskItem.RecurringOccurrenceOverride = recurringOverride;
+
+        if (existingOverride == null)
```
sol-wizard, line 75: **Major — override type/RecurringTaskId rewritten unconditionally.**

```csharp
recurringOverride.TaskItem = taskItem;
recurringOverride.RecurringTaskId = template.Id;
recurringOverride.OverrideType = RecurringOccurrenceOverrideType.Materialized;
recurringOverride.UpdatedAt = DateTime.UtcNow;
```

A couple of things tangled together here:

1. We've already early-returned for `Skipped`/`Detached` and for "existing override that already has a `TaskItem`". The only remaining cases are (a) brand-new override and (b) an existing override that's `Materialized` or `Modified` but somehow has no `TaskItem` attached. If case (b) ever happens (partial save, race, or a future code path that splits creation), this silently demotes `Modified → Materialized` and re-points it at the *current* template — discarding the user's custom edits.
2. The object initializer above (`new RecurringOccurrenceOverride { ... RecurringTaskId = template.Id, OverrideType = Materialized, ... }`) already sets these fields for case (a), so lines 60-61 are redundant in that path.

Suggest guarding the rewrite with `if (existingOverride == null)` so it's purely the new-override path, and turning case (b) into either a no-op (preserve `Modified`) or a logged assertion.

blotztask-api/Modules/Tasks/Commands/RecurringTasks/DeleteRecurringOccurrence.cs
```diff
@@ -0,0 +1,193 @@
...
+        recurringOverride.UpdatedAt = DateTime.UtcNow;
+    }
+
+    private async Task DeleteFutureOccurrences(
+        RecurringTask template,
+        DateOnly occurrenceDate,
+        Guid userId,
+        CancellationToken ct)
+    {
+        template.EndDate = occurrenceDate.AddDays(-1);
```
sol-wizard: **Real — edge case: `DeleteFuture` from the template's first occurrence leaves a "valid" template with `EndDate < StartDate`.**

If `occurrenceDate == template.StartDate`, this sets `EndDate = StartDate - 1` while leaving `IsActive = true`. `IsOccurrenceOn` correctly returns `false` for every date afterwards, so user-facing behavior is OK, but the row is in an invalid shape (`StartDate > EndDate`, still active) and the series can survive with no live versions.

Consider also flipping `IsActive = false` (or, if there are no remaining surviving overrides on the series, marking `Series.IsDeleted = true`) in this case so the data state matches the intent.

blotztask-api/Modules/Tasks/Services/RecurringTaskGeneratorService.cs
```diff
@@ -10,10 +11,70 @@ public class RecurringTaskGeneratorService
...
+            current = occurrence.Value;
+
+            var next = NextOccurrence(template, occurrence.Value);
+            if (next == null || next.Value <= occurrence.Value)
+            {
+                break;
+            }
+
+            occurrence = next;
+        }
```
sol-wizard, line 65: **Real — `GetCurrentOccurrenceDate` is O(occurrences-since-StartDate).**

This walks forward from `template.StartDate` one occurrence at a time. Called once per recurring deadline template from `GetAllDdlTasks` — for a daily template that's been around for 5 years that's ~1825 `NextOccurrence` calls per series, per `/api/deadline/all` request.

`Daily` has a closed form (essentially the inverse of `FirstDailyOccurrence` clamped to today), and weekly/monthly/yearly can step backward from the cutoff instead of forward from start. Not a blocker, but worth at least a TODO/comment so it's not forgotten; performance will degrade silently over time.

blotztask-api/Modules/Tasks/Commands/RecurringTasks/UpdateRecurringTaskFuture.cs
```diff
@@ -0,0 +1,434 @@
...
+
+        if (pattern.DayOfMonth is < 1 or > 31)
+        {
+            throw new ValidationException("DayOfMonth must be between 1 and 31.");
+        }
+
+        if (pattern.Frequency == RecurrenceFrequency.Weekly
+            && (pattern.DaysOfWeek == null || pattern.DaysOfWeek == (int)WeeklyDayFlags.None))
+        {
+            throw new ValidationException("DaysOfWeek is required for weekly recurring tasks.");
```
sol-wizard, line 288: **Real — validation asymmetry with `CreateRecurringTask`.**

`CreateRecurringTaskCommandHandler.ValidateRecurrence` enforces:

```csharp
if (details.Frequency != RecurrenceFrequency.Weekly && details.DaysOfWeek != null)
    throw new ValidationException("DaysOfWeek can only be set for weekly...");
if (details.Frequency != RecurrenceFrequency.Monthly && details.DayOfMonth != null)
    throw new ValidationException("DayOfMonth can only be set for monthly...");
```

This handler skips those checks. Combined with `BuildPattern` silently dropping `DaysOfWeek`/`DayOfMonth` when the new frequency doesn't match, a caller can send `{Frequency: Daily, DayOfMonth: 15}` and it'll be accepted-and-ignored. Not a security issue, just a confusing inconsistency for future maintainers/clients — happy if you add the symmetric guards here.

blotztask-api/Modules/Tasks/Queries/Deadlines/GetAllDdlTasks.cs
```diff
@@ -44,9 +56,101 @@ public async Task<List<DeadlineTaskDto>> Handle(GetAllDdlTasksQuery query, Cance
...
+        var recurringOverrideMap = await db.RecurringOccurrenceOverrides
+            .AsNoTracking()
+            .Include(o => o.TaskItem)
+            .ThenInclude(t => t!.Deadline)
+            .Include(o => o.TaskItem)
+            .ThenInclude(t => t!.Label)
+            .Where(o => o.Series.UserId == query.UserId
+                && currentSeriesIds.Contains(o.SeriesId)
+                && currentOccurrenceDates.Contains(o.OccurrenceDate))
+            .ToDictionaryAsync(o => new RecurringSeriesOccurrenceKey(o.SeriesId, o.OccurrenceDate), ct);
```
sol-wizard, line 102: **Real (lower-priority) — Cartesian filter on override lookup.**

The two `Contains` filters effectively become `SeriesId IN (...) AND OccurrenceDate IN (...)` in SQL, which is a cross product. Functionally fine because the in-memory dictionary keys by the exact `(SeriesId, OccurrenceDate)` tuple, but with N series you'll fetch up to N² rows.

For a heavy user (say, 50 recurring deadline templates) this is at worst ~2500 rows and you only need 50. Not worth refactoring unless this query shows up in traces — but if you want a tight fix later, one option is to build a `Dictionary<(int, DateOnly), ...>` of expected keys, send a list of `SeriesId`s and a `(min, max)` date window for the IN clause, and filter the rest client-side.

blotztask-mobile/src/feature/task-details/screens/task-details-screen.tsx
```diff
@@ -16,41 +16,101 @@ import {
...
+    };
+
+    if (hasTaskItemId(selectedTask)) {
+      await updateTask({
+        taskId: selectedTask.id,
+        dto,
+      });
+      return;
+    }
+
```
sol-wizard, line 101: nit (feel free to ignore) — `handleUpdateDescription` reads `recurringTaskId`, which is `const`-declared further down in the component body (after the `!selectedTask` early return). It works at runtime because the closure is only ever invoked after render commits, but it reads like a TDZ accident waiting to happen — moving the `const recurringTaskId = getRecurringTaskId(selectedTask)` above the handler (or passing it in as an arg) would make the ordering self-evident.

sol-wizard (commented):
Nice work — series + override + materializer is a clean way to model this and the new test coverage on the commands is substantial.

Most of my comments are inline. A few high-level points worth raising here:

**Worth a look before merge**
- Mobile silently flattens any `interval ≠ 1` (or `interval = 2` for biweekly) and any multi-day weekly pattern on the "edit future" flow. UI can't *create* those today, but the DTO already carries them and the backend supports them, so the first time anyone introduces a "custom" recurrence picker (or another client creates a task), edits will start corrupting rules. See comment on `task-edit-screen.tsx` / `mapTaskRecurrenceToRecurringPattern`.
- `Detached` recurring task items are reported as `MaterializedRecurringOccurrence` by `GetTasksByDate` but `NormalTaskItem` by `GetAllDdlTasks` — the mobile delete/edit flows key off this, so on the calendar a detached task incorrectly routes through the "this occurrence / future occurrences" sheet.
- The materializer's "I found an existing override without a TaskItem" branch unconditionally rewrites the override type to `Materialized` and re-points `RecurringTaskId` at the current template. Today that branch is only reachable for brand-new overrides, but if future code ever creates an override before the TaskItem, this would silently demote `Modified → Materialized`.

**Nice-to-have**
- `DeleteFutureOccurrences` leaves an `IsActive=true` template with `EndDate < StartDate` when called on the first occurrence — functional but the row is in an invalid shape.
- `GetCurrentOccurrenceDate` is O(occurrences-since-start); for long-running daily templates this'll get noticeable.
- `UpdateRecurringTaskFuture.ValidateRecurrence` is laxer than `CreateRecurringTask.ValidateRecurrence` — easy to bring into sync.
- `GetAllDdlTasks` override lookup is a Cartesian filter; fine at current scale.

**Things I checked and was happy with**
- Recurrence math for Daily / Weekly / Monthly / Yearly including leap-year and short-month skips.
- The `(SeriesId, OccurrenceDate)` unique index + `DbUpdateException` retry pattern in the materializer.
- `EditTask` flipping the override to `Modified` for materialized recurring task items — keeps the model consistent.
- Test coverage on `CreateRecurringTask`, `DeleteRecurringOccurrence`, the materializer, and `GetAllDdlTasks` is solid.

blotztask-api/Modules/Tasks/Queries/Deadlines/GetAllDdlTasks.cs
Xu-create-ops: Fixing this issue is quite complicated; it's just a performance issue, so we'll optimize it if we encounter any problems later on.

blotztask-api/Modules/Tasks/Domain/Entities/RecurringTask.cs
```diff
@@ -50,3 +64,29 @@ public class RecurringTask
     public DateTime CreatedAt { get; set; }
     public DateTime UpdatedAt { get; set; }
 }
+
+public class RecurringTaskSeries
```
sol-wizard: make this a new entities files so it is easier to navigate

blotztask-api/Modules/Tasks/Domain/Entities/RecurringTask.cs
```diff
@@ -50,3 +64,29 @@ public class RecurringTask
...
+    public required Guid UserId { get; set; }
+    public AppUser User { get; set; } = null!;
+    public bool IsDeleted { get; set; }
+    public DateTime CreatedAt { get; set; }
+    public DateTime UpdatedAt { get; set; }
+    public ICollection<RecurringTask> Versions { get; set; } = new List<RecurringTask>();
+    public ICollection<RecurringOccurrenceOverride> Overrides { get; set; } = new List<RecurringOccurrenceOverride>();
+}
+
+public class RecurringOccurrenceOverride
```
sol-wizard: make this a new entities files so it is easier to navigate

blotztask-api/Modules/Tasks/Queries/Deadlines/GetAllDdlTasks.cs
```diff
@@ -44,9 +56,101 @@ public async Task<List<DeadlineTaskDto>> Handle(GetAllDdlTasksQuery query, Cance
                     }
                     : null
             })
+            .ToListAsync(ct);
+
+        var recurringDeadlineTemplates = await db.RecurringTasks
```
sol-wizard, line 61: Add a todo mention this PBI https://github.com/Blotz-Org/Blotz-Task-App-Private/issues/1417

blotztask-api/Infrastructure/Data/Configurations/RecurringTaskConfiguration.cs
```diff
@@ -8,7 +8,20 @@ public class RecurringTaskConfiguration : IEntityTypeConfiguration<RecurringTask
 {
     public void Configure(EntityTypeBuilder<RecurringTask> builder)
     {
-        builder.ToTable("RecurringTasks");
+        builder.ToTable("RecurringTasks", t =>
```
sol-wizard, line 11: maybe this constraint can move to api layer validation

sol-wizard (approved):

blotztask-mobile/src/feature/task-details/screens/task-details-screen.tsx
```diff
@@ -16,41 +16,102 @@ import {
...
+  const taskId = params.taskId ? Number(params.taskId) : null;
+  const queryClient = useQueryClient();
+  const virtualTaskDetailQueryKey = virtualTaskDetailKeys.byKey(
+    params.virtualTaskCacheKey ?? "__missing__",
+  );
+  const { data: virtualTask } = useQuery<TaskDetailDTO | undefined>({
+    queryKey: virtualTaskDetailQueryKey,
+    queryFn: () => queryClient.getQueryData<TaskDetailDTO>(virtualTaskDetailQueryKey),
+    enabled: false,
+  });
```
sol-wizard, line 67: maybe need Nicole to review this part

sol-wizard (changes requested):

sol-wizard (approved):

PR #422, opened 2026-06-14
Input interaction

AlexisEvan:
## Release note
Retain the voice input page as initial. Add a keyboard toggle button; clicking it will automatically bring up the keyboard. Clicking the microphone button again will return to voice input.

> "none"

**Status:**


AlexisEvan:
截屏2026-06-14 下午6 31 34

截屏2026-06-14 下午6 32 15

blotztask-mobile/src/i18n/locales/en/ai-task-generate.json
```diff
@@ -6,6 +6,7 @@
     "speak": "Speak",
     "send": "Send",
     "tapToSpeak": "Tap to speak",
```
nxn-nicole: there is some language translation you no longer need, you can delete it

blotztask-mobile/src/feature/ai-task-generate/screens/ai-task-sheet-screen.tsx
```diff
@@ -194,49 +215,78 @@ export default function AiTaskSheetScreen() {
...
-                      } else {
-                        void handleMicPressOut();
-                      }
-                    }}
-                    onLongPress={() => {
-                      longPressTriggered.current = true;
-                    }}
-                    delayLongPress={1000}
+                    onPress={inputMode === "voice" ? handleSwitchToText : handleSwitchToVoice}
+                    accessibilityLabel={
```
nxn-nicole: do you know what accessibilityLabel is used for? i don't think you need accessibilityLabel

blotztask-mobile/src/feature/ai-task-generate/screens/ai-task-sheet-screen.tsx
```diff
@@ -77,6 +80,14 @@ export default function AiTaskSheetScreen() {
     });
   }, []);
 
+  // Focus the text field after it mounts so the keyboard reliably opens in this modal sheet
+  // (autoFocus is unreliable here — it can fire before the modal animation settles).
+  useEffect(() => {
```
nxn-nicole: Did you do a test here? I tested on my own device. It looks like just adding autofocus=true on the text input will achieve the same effect.

nxn-nicole (changes requested):

blotztask-mobile/src/feature/ai-task-generate/screens/ai-task-sheet-screen.tsx
AlexisEvan: accessibilityLabel is for accessibility . I added it because I saw it elsewhere and assumed it was expected here. I've removed it now since it's not needed.

blotztask-mobile/src/feature/ai-task-generate/screens/ai-task-sheet-screen.tsx
AlexisEvan: I used ref + manual focus because the keyboard occasionally failed to open in my own testing, so I went with what seemed more stable. Happy to use autoFocus since it's also reliable on device.

blotztask-mobile/src/i18n/locales/en/ai-task-generate.json
```diff
@@ -3,13 +3,7 @@
...
   },
   "buttons": {
-    "speak": "Speak",
-    "send": "Send",
-    "tapToSpeak": "Tap to speak",
-    "generateTask": "+ Generate Task",
-    "addToTasks": "Add to tasks",
-    "addToNotes": "Add to notes",
-    "addAll": "Add all",
+    "holdToTalk": "Hold to Talk",
```
sol-wizard, line 6: **Minor — unrelated cleanup.**

Removing `speak`, `send`, `tapToSpeak`, `generateTask`, `addToTasks`, `addToNotes`, `addAll` is dead-key cleanup unrelated to the voice/text toggle. I confirmed via grep that none are referenced elsewhere, so the deletion is safe — but ideally this would be a separate PR (or at least called out in the empty `## Summary` section) so reviewers / git blame don't have to detangle UI work from i18n housekeeping.

sol-wizard (commented):
Clean, focused UI change. The mode-toggle pattern works correctly — voice default matches the release note, the toggle is properly disabled mid-recording / mid-generation, `autoFocus` re-fires on remount so the keyboard pops reliably on switch, and `textInput` state is preserved across switches. The removed i18n keys are confirmed dead.

Three minor / speculative inline notes:

1. The new mode-toggle Pressable is missing an `accessibilityLabel` (the existing dismiss button has one).
2. `handleSwitchToVoice` doesn't reset `isHoldHintVisible`, which could latch the "hold longer" hint across mode switches in a niche edge case.
3. The 7 removed i18n keys are unrelated cleanup — would have been nicer as a separate PR, or at least called out in the (empty) Summary.

Nothing blocking. Nice work.

nxn-nicole (approved):

PR #428, opened 2026-06-17
Feature/1410 store user timezone

Guotai812:
## Summary

Stores the user's device timezone after a successful Auth0 login so the backend has it available for future features (server-side reminders, recurring task scheduling, AI date interpretation).

**Backend:**
- Added nullable `Timezone` column to `AppUser` with EF migration
- Extended `PUT /api/User` to accept and validate `timezone` (IANA/Windows timezone ID) using `TimeZoneInfo.FindSystemTimeZoneById`
- `GET /api/User` now returns `timezone` in the profile response

**Mobile:**
- After Auth0 `authorize()` succeeds and tokens are saved, reads device timezone via `Intl.DateTimeFormat().resolvedOptions().timeZone` and calls `PUT /api/User` with it
- Failure is non-blocking — login completes regardless of whether the timezone sync succeeds

## Release note

> none

**Status:**


blotztask-api/Modules/Users/Commands/UpdateUserProfile.cs
```diff
@@ -43,6 +45,14 @@ public async Task<string> Handle(UpdateUserProfileCommand profileCommand, Cancel
             hasChange = true;
         }
 
+        if (!string.IsNullOrWhiteSpace(profileCommand.Timezone))
+        {
+            try { TimeZoneInfo.FindSystemTimeZoneById(profileCommand.Timezone); }
+            catch (Exception) { throw new ValidationException($"Invalid timezone: {profileCommand.Timezone}"); }
```
sol-wizard: **Major:** `catch (Exception)` is too broad. `FindSystemTimeZoneById` only throws `TimeZoneNotFoundException` or `InvalidTimeZoneException` — catching all `Exception` masks unrelated failures (ICU/globalization, OOM, etc.) as "Invalid timezone" and returns a misleading 400.

Suggested:
```csharp
try { TimeZoneInfo.FindSystemTimeZoneById(profileCommand.Timezone); }
catch (Exception ex) when (ex is TimeZoneNotFoundException or InvalidTimeZoneException)
{
    throw new ValidationException($"Invalid timezone: {profileCommand.Timezone}");
}
```

blotztask-api/Modules/Users/Domain/AppUser.cs
```diff
@@ -16,4 +16,6 @@ public class AppUser
     public DateTime? LoginAt { get; set; }
     public PomodoroSetting PomodoroSetting { get; set; } = null!;
     public bool IsOnboarded { get; set; } = false;
+    
+    public string? Timezone { get; set; }
 }
```
sol-wizard, line 23: **Minor:** the migration stores this as `nvarchar(max)`, but IANA IDs max ~32 chars and Windows IDs ~80. Consider `[MaxLength(64)]` (or `100`) — more efficient storage and an implicit input sanity bound. Easy to do now while the column is empty.

blotztask-api/Infrastructure/Data/Migrations/20260617032029_AddTimezoneToAppUser.cs
```diff
@@ -0,0 +1,36 @@
...
+                table: "AppUsers",
+                type: "nvarchar(max)",
+                nullable: true);
+
+            migrationBuilder.UpdateData(
+                table: "AppUsers",
+                keyColumn: "Id",
+                keyValue: new Guid("718ccb8f-ce52-4e51-8cfe-2a44cdca77d1"),
+                column: "Timezone",
+                value: null);
```
sol-wizard: **Minor / nit:** this `UpdateData` setting `Timezone = null` for the seeded user is redundant — the column is nullable so it defaults to null. EF auto-gen artifact from seed data; harmless but you can delete it for a cleaner migration.

sol-wizard (commented):
**AI-generated PR review.** Treat these as suggestions — please use your own judgement and verify before changing anything.

The PR is small and well-scoped, and the description matches the diff. Two notes that didn't fit inline:

**Cross-platform timezone ID compatibility (please confirm prod runtime).** Mobile sends IANA IDs (`America/Los_Angeles`). `TimeZoneInfo.FindSystemTimeZoneById` accepts IANA on Linux/macOS, and accepts both IANA + Windows IDs on .NET 6+ *only when ICU is available* (globalization-invariant mode disabled). If the API is hosted on Linux App Service → fine. If on Windows or in an invariant-globalization container, IANA IDs will be rejected and every mobile login will log a 400. Worth a quick check before merging.

**Speculative — feel free to ignore:** timezone is only captured at first login (`get-started-button`). If a user travels/relocates, the stored value goes stale and any future server-side reminders / AI date interpretation will use the wrong zone. Probably out of scope for this PR — just flagging that a refresh point (app foreground, token refresh) will be needed before those features ship.

sol-wizard (approved):

PR #430, opened 2026-06-20
fix relogin issue using auth0 lib packages for refresh token

sol-wizard:
## Summary

Made the Auth0 SDK credentials manager the single source of truth for tokens and removed the duplicate expo-secure-store copy. The SDK now owns the refresh token and handles rotation internally, so the two stores can't collide anymore.

## Release note

Users stay logged in and no longer get randomly signed out. Tokens refresh silently in the background; the only sign-in needed is the normal one when a session genuinely reaches its expiry which is 90 days.

**Status:**

### DIFF
--- a/blotztask-mobile/src/feature/ai-task-generate/services/ai-task-generator-signalr-service.ts
+++ b/blotztask-mobile/src/feature/ai-task-generate/services/ai-task-generator-signalr-service.ts
@@ -9,17 +9,19 @@ const SIGNALR_HUBS_CHAT = `${config.API_BASE_URL}/ai-task-generate-chathub`;
 export const signalRService = {
   createConnection: async () => {
     console.log("Attempting to create SignalR connection to:", SIGNALR_HUBS_CHAT);
-    const token = await getAuthToken();
-    if (!token) {
-      throw new Error("No valid Auth0 credentials available.");
-    }
 
     const timeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
     const url = `${SIGNALR_HUBS_CHAT}?timeZone=${encodeURIComponent(timeZone)}`;
 
     const connection = new signalR.HubConnectionBuilder()
       .withUrl(url, {
-        accessTokenFactory: () => token,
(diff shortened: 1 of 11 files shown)


blotztask-mobile/src/shared/services/api/token-manager.ts
```diff
@@ -1,60 +1,42 @@
...
+  }
 }
 
-export async function getRefreshToken(): Promise<string | null> {
-  return await SecureStore.getItemAsync(REFRESH_TOKEN_KEY);
+/**
+ * Forces a refresh that bypasses the cached access token. Used to retry once
+ * after a 401 in case the access token was stale.
+ */
+export async function forceRefreshAuthToken(): Promise<string | null> {
```
Xu-create-ops: Consider concurrent access here. Concurrent calls to forceRefreshAuthToken after token expiration may revoke the session.

blotztask-mobile/src/feature/ai-task-generate/services/ai-task-generator-signalr-service.ts
```diff
@@ -10,10 +9,9 @@ const SIGNALR_HUBS_CHAT = `${config.API_BASE_URL}/ai-task-generate-chathub`;
 export const signalRService = {
   createConnection: async () => {
     console.log("Attempting to create SignalR connection to:", SIGNALR_HUBS_CHAT);
-    const token = await SecureStore.getItemAsync(AUTH_TOKEN_KEY);
-    // TODO: Add error handling
+    const token = await getAuthToken();
```
Xu-create-ops: The current implementation fetches the token only once when the connection is created and then uses it via accessTokenFactory: () => token.  Consider moving getAuthToken into accessTokenFactory to ensure a fresh token is used and avoid issues caused by stale tokens.（e.g. accessTokenFactory: async () => {
  const token = await getAuthToken();
  if (!token) throw new Error("No valid Auth0 credentials available.");
  return token;
}）

Xu-create-ops (approved):

PR #435, opened 2026-06-23
group date format in one file

LiChenyang-GZ:
## Summary

Centralized localized date display formatting in a shared mobile utility and replaced duplicated date-fns locale/pattern logic across task add/edit, calendar, badge, and review screens.

This is a refactor only; existing English and Chinese date display formats are intended to stay the same.

## Release note

> none

**Status:**

### DIFF
--- a/blotztask-mobile/src/feature/calendar/util/date-formatter.tsx
+++ b/blotztask-mobile/src/feature/calendar/util/date-formatter.tsx
@@ -57,7 +57,7 @@ export const formatCalendarDate = (
 export const renderCalendarHeader = (date?: unknown) => {
   if (!date) return null;
   const dateObj = date instanceof Date ? date : new Date(date.toString());
-  const dateText = formatLocalizedDate(dateObj, "monthYear", i18n.language);
+  const dateText = formatLocalizedDate(dateObj, "monthYear");
   return (
     <Text
       style={{
@@ -69,15 +69,4 @@ export const renderCalendarHeader = (date?: unknown) => {
       {dateText}
     </Text>
   );
-};
-
-/**
- * Formats a date for the Monthly Calendar BottomSheet display
- * @param date - Date object or date string
- * @returns Formatted date string (English: Mon, 15 Apr 2026 | Chinese: 4月15日)
- */
-export const formatBottomSheetDate = (date?: unknown) => {
-  if (!date) return "";
-  const dateObj = date instanceof Date ? date : new Date(date.toString());
-  return formatLocalizedDate(dateObj, "dateWithWeekday", i18n.language);
 };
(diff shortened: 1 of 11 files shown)


blotztask-mobile/src/feature/calendar/util/date-formatter.tsx
```diff
@@ -80,11 +79,5 @@ export const renderCalendarHeader = (date?: unknown) => {
 export const formatBottomSheetDate = (date?: unknown) => {
```
nxn-nicole: Do you still need this formatBottomSheetDate function, or I feel like you can directly use the formatLocalizedDate function in MonthlyCalendarScreen. because formatBottomSheerDate just validates the date type.

blotztask-mobile/src/shared/util/localized-date-format.ts
```diff
@@ -0,0 +1,39 @@
...
+  },
+  yearMonth: {
+    en: "MMMM yyyy",
+    zh: "yyyy年M月",
+  },
+};
+
+type LocalizedDateFormatPreset = keyof typeof DATE_FORMAT_PATTERNS;
+
+export const formatLocalizedDate = (
```
nxn-nicole: import i18n from 'i18next';

export const formatLocalizedDate = (
  date: Date,
  dateFormat: LocalizedDateFormatPreset,
) => {
  const userLanguage = i18n.language.toLowerCase().startsWith("zh") ? "zh" : "en";
  // ...
};
Maybe you can get user language inside of these two functions, so you don't need to pass extra props from all the other places.

blotztask-mobile/src/shared/util/localized-date-format.ts
```diff
@@ -0,0 +1,39 @@
+import { format } from "date-fns";
+import { enUS, zhCN } from "date-fns/locale";
+
+const DATE_FORMAT_PATTERNS = {
```
nxn-nicole, line 5: Naming and Readability
The date format patterns here might be difficult to read elsewhere. For example, when reading "long date" in other parts of the code, it's hard to immediately understand that it corresponds to these two time formats.

Image

nxn-nicole (changes requested):

blotztask-mobile/src/feature/calendar/util/date-formatter.tsx
LiChenyang-GZ: ok

blotztask-mobile/src/shared/util/localized-date-format.ts
LiChenyang-GZ: ok

blotztask-mobile/src/shared/util/localized-date-format.ts
LiChenyang-GZ: ok

nxn-nicole (approved):

PR #436, opened 2026-06-23
feat(badge): add View and Share Reward actions to achievement popup

dksjl661:
## Summary

PR 1 of 2 for the Badge Reward popup enhancement. Adds footer actions to the badge achievement popup and the shareable badge image flow. Celebration effects (Lottie stars, haptics, sound) are deferred to PR 2.

- Add `View` button (temporary `console.log` placeholder, per spec — no navigation yet)
- Add `Share Reward` button that generates a branded badge image and opens the native share sheet
- New off-screen `BadgeShareCard` (`expo-linear-gradient` background for reliable iOS/Android capture) with badge icon, title, description, obtained date, and Blotz branding
- New `useBadgeShare` hook reusing the Monthly Report share flow (`react-native-view-shot` `captureRef` + `expo-sharing` `shareAsync`)
- Add en/zh i18n keys: `view`, `shareReward`, `sharing`, `shareError`

Language consistency: the share image renders the already-localized `name`/`description` from the badge DTO, and header/date use the `badge` i18n namespace, so it follows the user's language preference.

## Release note

> You can now share an earned badge as an image directly from the achievement popup.

**Status:**

- [ ] User-facing — announce it
- [x] Beta / partial — announce, but tagged as beta
- [ ] Hidden in production (feature-flagged / not enabled for users) — don't announce
- [ ] Internal only (refactor / infra / tests / CI / deps) — don't announce

## Test plan

- [ ] Trigger the badge popup (badge push notification, or seed SQL + complete an overdue task)
- [ ] Tap `View` → logs `Badge view pressed <id>`
- [ ] Tap `Share Reward` → native share sheet opens with a generated image containing icon, title, description, date, Blotz branding
- [ ] Switch language (zh/en) and confirm popup + share image text follow the language
- [ ] Verify on both iOS and Android

## Out of scope (PR 2)

- Celebration star Lottie animation, haptic feedback, and sound effect on popup appear
- Wiring `View` to real navigation

PR #437, opened 2026-06-23
feat(badge): add celebration effects to achievement popup

dksjl661:
## Summary

PR 2 of 2 for the Badge Reward popup enhancement. Adds celebration feedback that fires once each time the badge achievement popup appears (and re-triggers for the next queued badge).

Stacked on #436 (`feature/badge-reward-share-actions`) — review/merge that PR first; this PR is based on it so the diff stays scoped to the celebration effects.

- Overlay a confetti/star Lottie (`autoPlay`, no loop, non-interactive) keyed on `badgeId` so it replays per badge
- Play a one-shot reward sound via `expo-audio` `useAudioPlayer`
- Trigger a success haptic via `expo-haptics` (`notificationAsync(Success)`)
- Drive all three from a `useEffect` keyed on `badgeId`; audio and haptics degrade gracefully on unsupported devices, and audio failure never blocks the popup
- Add placeholder assets: `assets/animations/badge-celebration.json` + `assets/sounds/badge-reward.mp3`, registered in `assets.ts`

Note: Lottie and sound are free placeholder assets pending final designer assets.

## Release note

> Earning a badge now celebrates with a confetti animation, sound, and haptic feedback.

**Status:**

## Test plan

- [x] Trigger the badge popup (badge push notification, or seed SQL + complete an overdue task)
- [x] Confetti/star animation plays once when the popup appears
- [ ] One reward sound plays once; one success haptic fires once
- [x] Dismiss; the next queued badge re-triggers all effects
- [ ] Verify graceful degradation: device with haptics unsupported / silent mode still shows the popup normally
- [ ] Verify on both iOS and Android

Verified on iOS simulator: confetti renders over the popup and the View/Share Reward actions still work without crashing.


dksjl661:
Simple question, do i need to move all the badge into database? And sorry the mp3 cannot be tested in the comment
But I am sure it works

blotztask-mobile/src/feature/badge/components/badge-achievement-modal.tsx
```diff
@@ -20,11 +24,39 @@ export function BadgeAchievementModal({ badge, onDismiss }: BadgeAchievementModa
...
+      // Haptics unsupported on this device - degrade gracefully.
+    });
+
+    try {
+      rewardSound.seekTo(0);
+      rewardSound.play();
+    } catch {
+      // Audio failure must never block the popup.
+    }
+  }, [badgeId, rewardSound]);
```
nxn-nicole: Why do you need to add reward sound to the dependency? Will reward sound be changing in the middle?

blotztask-mobile/src/feature/badge/components/badge-achievement-modal.tsx
```diff
@@ -20,11 +24,39 @@ export function BadgeAchievementModal({ badge, onDismiss }: BadgeAchievementModa
...
+  const rewardSound = useAudioPlayer(ASSETS.badgeReward);
+
+  const badgeId = badge?.badgeId;
+
+  // Fire celebration feedback once per badge (re-fires for the next queued badge).
+  useEffect(() => {
+    if (badgeId == null) return;
+
+    Haptics.notificationAsync(Haptics.NotificationFeedbackType.Success).catch(() => {
+      // Haptics unsupported on this device - degrade gracefully.
```
nxn-nicole: Do you need a catch here? In which case the Haptics will not be supported and might cause an error？

nxn-nicole (changes requested):

blotztask-mobile/src/feature/badge/components/badge-achievement-modal.tsx
dksjl661: rewardSound is used inside the effect, so it needs to be included in the dependency array. useAudioPlayer can recreate the player if its source or configuration changes; including it ensures the effect uses the current player. The current source is static, so it will not cause extra runs in normal usage.

blotztask-mobile/src/feature/badge/components/badge-achievement-modal.tsx
dksjl661: notificationAsync can reject when the native Haptics module is unavailable (for examplee, an unsupported or improperly linked runtime). This feedback is optional, so the catch prevents an unhandled rejection while allowing the achievement popup to display normally.

PR #443, opened 2026-06-26
Firework feedback animation

lyx2-2:
## Release note

> _your one-line user-facing note here, or "none"_

**Status:**


lyx2-2:
https://github.com/user-attachments/assets/3aa7b895-76f1-477c-9a3e-e107ff25ec25

blotztask-mobile/src/feature/calendar/components/subtask-list.tsx
```diff
@@ -31,9 +31,13 @@ const SubtaskList = ({ task, progress }: Props) => {
     overflow: "hidden",
   }));
 
-  const handleToggleSubtask = (subtaskId: number) => {
+  const handleToggleSubtask = (subtask: SubtaskDTO) => {
     if (task.id == null) return;
```
dksjl661, line 35: Does it really need to pass the whole DTO just for passing a "wasdone"?

blotztask-mobile/src/feature/firework-animation/hooks/useFirework.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+type FireworkContextValue = Record<FireworkVariant, FireworkSlot>;
+
+const FireworkContext = createContext<FireworkContextValue | null>(null);
+
+function useFireworkSlot(): FireworkSlot {
+  const [playbackKey, setPlaybackKey] = useState(0);
+  const [visible, setVisible] = useState(false);
+
+  const play = useCallback(() => {
```
dksjl661: why we need usecallback? the cost is not such high i think. Correct me If I am wrong including the remaining usecallback and usememo

blotztask-mobile/src/feature/firework-animation/components/firework-overlay.tsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+function FireworkOverlay({ variant }: { variant: FireworkVariant }) {
+  const { width, height } = useWindowDimensions();
+  const { visible, playbackKey, dismiss } = useFirework()[variant];
+  const config = FIREWORK_OVERLAY_CONFIG[variant];
+  const animationSize = config.getSize(width, height);
+
+  return (
+    <Modal
```
dksjl661: If the firework is intended to be non-blocking visual feedback, `Modal` may still intercept the screen even with `pointerEvents="box-none"` inside it. Could we verify whether taps/gestures should pass through during the animation, or use an absolutely positioned overlay in the protected layout instead?

dksjl661 (approved):

PR #448, opened 2026-06-27
fix(gashapon): guarantee dispense after spin

dksjl661:
## Summary
- Guarantee a note is dispensed after a successful knob rotation, independent of device tilt.
- Remove the selected star from both the physics world and render entity map to avoid duplicate drops.
- Stabilize the dropped-star callback and empty notes search result to avoid render loops.
- Update gashapon help copy so shaking is optional and tilt is no longer required.

## Release note

> The gacha machine now reliably drops a Note after a successful spin without requiring phone tilt.

**Status:**

## Notes
- Backend `NotesController` debug changes were intentionally excluded because they were only needed to unblock a local simulator/API config issue and are unrelated to PBI 1442.

## Verification
- `npm run lint` passes with existing warnings only.
- `npm exec eslint -- src/feature/gashapon-machine/hooks/useGashaponMachineConfig.ts src/feature/gashapon-machine/screens/gashapon-machine-screen.tsx src/feature/notes/hooks/useNotesSearch.ts`
- `npm exec prettier -- --check src/feature/gashapon-machine/hooks/useGashaponMachineConfig.ts src/feature/gashapon-machine/screens/gashapon-machine-screen.tsx src/feature/notes/hooks/useNotesSearch.ts src/i18n/locales/en/notes.json src/i18n/locales/zh/notes.json`
- `npm exec tsc -- --noEmit --skipLibCheck --pretty false` fails on existing third-party/react-native-calendars type errors, `src/app/_layout.tsx` global.css side-effect import, and existing `src/feature/task-details/components/subtasks-editor.tsx` `SubtaskDTO.id` errors; no changed-file errors found.
- Manual iOS simulator verification: rotate the knob -> Reveal A New Task modal appears.


blotztask-mobile/src/feature/gashapon-machine/hooks/useGashaponMachineConfig.ts
```diff
@@ -45,16 +49,65 @@ export const useGashaponMachineConfig = ({
...
+    const selectedStarIndex =
+      availableStarIndexes[Math.floor(Math.random() * availableStarIndexes.length)];
+    const selectedStar = starsRef.current[selectedStarIndex];
+
+    droppedStarIndexRef.current = selectedStarIndex;
+    Matter.World.remove(world, selectedStar);
+    removeStarEntity(selectedStarIndex);
+    closeGate();
+    onStarDroppedRef.current(selectedStarIndex);
+  };
```
LiChenyang-GZ: Question: this now randomly selects any available star on release instead of using the physical star that reaches the drop sensor. If the user has moved one star near the outlet but `dispenseStar()` randomly selects another, the revealed note can differ from the visible star position. Is that intentional for guaranteed dispense, or should we preserve the physical drop as the source of truth and only add a fallback when no collision happens?

LiChenyang-GZ (changes requested):

blotztask-mobile/src/feature/gashapon-machine/hooks/useGashaponMachineConfig.ts
dksjl661: Good point. I agree we should preserve the physical drop as the source of truth.

The random selection here was meant to guarantee a dispense, but as you pointed out it can disconnect the visible star position from the revealed note. I’ll adjust this so release opens the gate and lets the drop sensor determine the dropped star first, then only use a random fallback if no collision happens within a short timeout.

dksjl661: The random selection is intentional. The notes are hidden inside the machine and the stars are not individually identifiable to the user, so there is no visible star-to-note mapping to preserve.

This PR is specifically meant to stop relying on the physical star reaching the drop sensor, because that made dispense unreliable. The machine should behave like a random draw once the spin threshold is met.

dksjl661:
Closing this PR. The intended behavior is being reconsidered separately.

PR #454, opened 2026-07-02
Create get badge by id endpoint

AlexisEvan:
## Summary

Add a new backend endpoint that returns a single badge's
details, localized to the user's language preference (defaults to English), and returns 404 for non-existent
badge IDs. Not yet consumed by any user-facing UI.

## Release note

none
>  "none"_

**Status:**

### DIFF
--- a/blotztask-api/Modules/Badges/Queries/GetBadgeById.cs
+++ b/blotztask-api/Modules/Badges/Queries/GetBadgeById.cs
@@ -1,0 +1,46 @@
+using BlotzTask.Infrastructure.Data;
+using BlotzTask.Modules.Users.Enums;
+using Microsoft.EntityFrameworkCore;
+using BlotzTask.Shared.Exceptions;
+
+namespace BlotzTask.Modules.Badges.Queries;
+
+public class GetBadgeByIdQuery
+{
+    public required Guid UserId { get; init; }
+    public required int BadgeId { get; init; }
+}
+
+public class GetBadgeByIdQueryHandler(BlotzTaskDbContext db, ILogger<GetBadgeByIdQueryHandler> logger)
+{
+    public async Task<BadgeByIdItemDto> Handle(GetBadgeByIdQuery query, CancellationToken ct = default)
+    {
+        logger.LogInformation("Fetching badges for UserId {UserId}", query.UserId);
+
+        var preference = await db.UserPreferences.FindAsync(query.UserId, ct);
+        var language = preference?.PreferredLanguage ?? Language.En;
+        var badge = await db.Badges.FindAsync(query.BadgeId, ct);
+
+        if (badge == null)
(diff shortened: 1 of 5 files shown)


blotztask-mobile/src/feature/badge/models/badge-by-id-dto.ts
```diff
@@ -0,0 +1,7 @@
+export interface BadgedDetailDTO {
```
dksjl661: nit: `BadgedDetailDTO` looks like a typo. Please rename it to `BadgeDetailDTO` so future imports match the badge detail concept clearly.

dksjl661:
For future endpoint work, please include a successful Swagger screenshot, curl output, or response log in the PR/issue. Code review can check the implementation shape, but the successful request evidence is what shows the feature actually works locally.

dksjl661 (commented):
please treat this as a suggestion and verify before acting.\n\nNo blocking issues found against Blotz-Task-App-Private#1439. The endpoint returns localized name/description, icon URL, category, and invalid badge IDs flow through NotFoundException as 404. I also verified `dotnet restore blotztask-api/BlotzTask.csproj && dotnet build blotztask-api/BlotzTask.csproj --no-restore` on the PR worktree; build passed with existing warnings.\n\nminor: `BadgedDetailDTO` in `blotztask-mobile/src/feature/badge/models/badge-by-id-dto.ts` looks like a typo; consider `BadgeDetailDTO` before it gets imported elsewhere.

blotztask-api/Modules/Badges/Queries/GetBadgeById.cs
```diff
@@ -0,0 +1,46 @@
...
+            IconUrl = badge.IconUrl,
+            Category = badge.Category.ToString(),
+        };
+    }
+}
+
+public class BadgeByIdItemDto
+{
+    public int Id { get; set; }
+    public required string Title { get; set; }
```
LiChenyang-GZ, line 42: could you please change the Title to Name(also other places in badge not only this one)? Cause the domain is nameZh/nameEn, PBI also require to return name, also we have BadgeNotificationDTO use name instead of badge. This is an API contract naming risk.

LiChenyang-GZ (changes requested):

AlexisEvan:
截屏2026-07-04 下午2 52 09

AlexisEvan:
截屏2026-07-04 下午2 53 39

blotztask-api/Modules/Badges/Queries/GetBadgeById.cs
AlexisEvan: One concern: `Title` is also used in the already-merged list endpoint
(`GetUserBadges` + its frontend), and renaming the field is an API-contract
change, so front and back have to change together.

dksjl661 (approved):

LiChenyang-GZ (approved):

PR #456, opened 2026-07-03
Add backend support for persistent notes

dksjl661:
## Summary

Adds backend support for persistent Notes. Persistent Notes can now be created, updated, listed, and converted to Tasks without deleting the original Note. Normal Notes keep the existing one-time conversion behavior.

## Release note

> none

**Status:**

## Changes

- Add `IsPersistent` to the Note entity and EF configuration with default `false`.
- Accept `isPersistent` in note create/update requests.
- Return `isPersistent` in Note DTOs and search/list responses.
- Keep persistent Notes when converting to Tasks; delete normal Notes as before.

## Migration

This requires an EF migration from `blotztask-api/`:

```bash
dotnet ef migrations add AddIsPersistentToNotes
dotnet ef database update
```

## Validation

- `dotnet build blotztask-api/BlotzTask.csproj --no-restore`

Tests were intentionally removed per follow-up request.

### DIFF
--- a/blotztask-api/Modules/Notes/Commands/CreateNote.cs
+++ b/blotztask-api/Modules/Notes/Commands/CreateNote.cs
@@ -9,7 +9,7 @@ public class CreateNoteCommand
 {
   [Required] public Guid UserId { get; set; }
   [Required] public string Text { get; set; } = string.Empty;
-  public bool IsPersistent { get; set; }
+  [Required] public bool IsPersistent { get; set; }
 }
 public class CreateNoteCommandHandler(BlotzTaskDbContext db, ILogger<CreateNoteCommandHandler> logger)
 {
--- a/blotztask-api/Modules/Notes/Commands/CreateNote.cs
+++ b/blotztask-api/Modules/Notes/Commands/CreateNote.cs
@@ -9,7 +9,7 @@ public class CreateNoteCommand
 {
   [Required] public Guid UserId { get; set; }
   [Required] public string Text { get; set; } = string.Empty;
-  [Required] public bool IsPersistent { get; set; }
+  public bool? IsPersistent { get; set; }
 }
 public class CreateNoteCommandHandler(BlotzTaskDbContext db, ILogger<CreateNoteCommandHandler> logger)
 {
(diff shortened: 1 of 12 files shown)


blotztask-api/Modules/Notes/Commands/CreateNote.cs
```diff
@@ -9,6 +9,7 @@ public class CreateNoteCommand
 {
   [Required] public Guid UserId { get; set; }
   [Required] public string Text { get; set; } = string.Empty;
+  public bool IsPersistent { get; set; }
```
LiChenyang-GZ: why not required

LiChenyang-GZ (changes requested):

blotztask-api/Modules/Notes/Commands/UpdateNote.cs
```diff
@@ -13,6 +13,7 @@ public class UpdateNoteCommand
   [Required] public required Guid NoteId { get; set; }
   [Required] public required string Text { get; set; }
   [Required] public required Guid UserId { get; set; }
+  [Required] public bool IsPersistent { get; set; }
```
LiChenyang-GZ: required

blotztask-api/Modules/Notes/Domain/Note.cs
```diff
@@ -11,10 +11,11 @@ public class Note
   public string Text { get; set; } = string.Empty;
   public DateTime CreatedAt { get; set; }
   public DateTime UpdatedAt { get; set; }
+  public bool IsPersistent { get; set; }
```
LiChenyang-GZ, line 14: pls add IsPersistent in default note when user onboarding

LiChenyang-GZ (changes requested):

blotztask-api/Modules/Notes/Controllers/NotesController.cs
```diff
@@ -26,7 +26,8 @@ public async Task<NoteDto> CreateNote([FromBody] NoteRequestDto request, Cancell
         var command = new CreateNoteCommand
         {
             UserId = userId,
-            Text = request.Text
+            Text = request.Text,
+            IsPersistent = request.IsPersistent ?? false
```
LiChenyang-GZ: pls move to handler to deal with this

LiChenyang-GZ (changes requested):

LiChenyang-GZ (approved):

PR #468, opened 2026-07-12
assign display order when awarding badges

lyx2-2:
## Summary

Assign a display order to newly awarded badges based on the user's current
badge count (0/1/2, null beyond three). Backend logic for #1440.

## Release note

> none

**Status:**


dksjl661 (approved):
that's insane mate

LiChenyang-GZ (approved):

PR #470, opened 2026-07-12
fix(ai): preserve drafts when quota is exceeded

dksjl661:
## Summary

- preserve existing AI-generated task and note drafts when the backend reports `QuotaExceeded`
- show a dedicated quota-limit message in English and Chinese
- keep the existing clearing behavior for all other generation errors

## Root cause

The generation error handler cleared all streamed drafts for every error code. Quota checks happen before streaming starts, so a quota rejection erased drafts from earlier turns even though the rejected turn produced nothing.

## Validation

- `npm ci`
- `npx eslint . --ext .ts,.tsx,.js` (0 errors; existing warnings only)
- API restore, Release build, publish, and EF migration script generation
- backend test run attempted; local SQL Server Testcontainers exhausted Docker memory, so GitHub Actions remains the authoritative full-suite run

## Release note

> AI-generated drafts are now kept when you reach your AI usage limit.

**Status:**

### DIFF
--- a/blotztask-mobile/src/feature/ai-task-generate/hooks/useAiTaskGenerator.ts
+++ b/blotztask-mobile/src/feature/ai-task-generate/hooks/useAiTaskGenerator.ts
@@ -21,6 +21,7 @@ const ERROR_CODE_TO_I18N_KEY: Record<string, string> = {
   BlockedByContentFilter: "errors.contentFilter",
   Canceled: "errors.canceled",
   NoTasksExtracted: "errors.noTasksExtracted",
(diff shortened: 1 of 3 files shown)


dksjl661:
PBI: https://github.com/Blotz-Org/Blotz-Task-App-Private/issues/1483

This PR implements PBI #1483: preserve existing generated task and note drafts when AI quota is exceeded, show a clear localized quota message, and keep existing behavior for other generation errors.

Guotai812 (approved):

PR #472, opened 2026-07-13
fix typing input accidentally submit

AlexisEvan:
## Summary

- Cards stay visible when the keyboard opens.
- the confirm (✓) button is hidden while the keyboard is up.
- The card list reserves the keyboard's height, so it ends above the floating input bar instead of being covered by it.
- When there are multiple cards and the keyboard pops up, user can swipe to view all the cards.

## Release note

none

**Status:**

### DIFF
--- a/blotztask-mobile/src/feature/ai-task-generate/screens/ai-task-sheet-screen.tsx
+++ b/blotztask-mobile/src/feature/ai-task-generate/screens/ai-task-sheet-screen.tsx
@@ -8,7 +8,12 @@ import {
   Keyboard,
   Platform,
 } from "react-native";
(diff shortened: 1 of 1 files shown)


AlexisEvan:
The first version reserved the list's bottom space with a `paddingBottom` driven by React state. During testing, whenever the keyboard popped up or retracted, the page showed a noticeable stutter.

So this version use `useReanimatedKeyboardAnimation()` instead of React state. Which makes the keyboard show/hide visibly smoother.

AlexisEvan:
截屏2026-07-14 上午12 04 04

dksjl661 (approved):
LGTM

PR #475, opened 2026-07-17
add invitation system

Moomin928:
---Summary

Adds a referral code system:

Backend:

- New ReferralCodes and Referrals tables (replaces the old InviteRedemptions table and InviteCode column on AppUsers)
- GET /api/referral/my-code — lazily ensures and returns the user's unique referral code (self-healing: generates one on first access if it's missing, regardless of whether the Auth0 login webhook already created it)
- POST /api/referral/redeem — validates and records a redemption (only allowed before onboarding is complete; prevents self-use and double redemption)
- Codes are derived from each row's own auto-increment Id via Sqids (guaranteed unique by construction, no random-collision retries needed), using an uppercase-only alphabet that excludes visually ambiguous characters
- Error handling: 404 code/user not found, 400 self-referral, 403 already onboarded, 409 already redeemed

Frontend:

- Invite code redemption added to the onboarding flow, plus an "Invite Friends" entry in Settings showing the user's own code
- One-tap copy for the user's own code
- Input field (up to 12 characters) to redeem another user's code
- Localised error messages (EN/ZH) per error type

Design note: redemption is intentionally limited to the onboarding window — this is a deliberate anti-abuse decision, not an oversight. Consequences: existing (already-onboarded) users cannot participate, and tapping Skip on any onboarding step permanently forfeits the ability to redeem, since it marks the user as onboarded. There is no redeem entry point outside onboarding by design.

##Release note

 You can now invite friends with your own referral code.

Status:

---


Moomin928:
https://github.com/Blotz-Org/Blotz-Task-App-Private/issues/1464

Moomin928:
Simulator Screenshot - iPhone 17 Pro Max - 2026-07-17 at 00 31 20
Simulator Screenshot - iPhone 17 Pro Max - 2026-07-17 at 00 31 25

Moomin928:
Simulator Screenshot - iPhone 17 Pro Max - 2026-07-27 at 18 22 53
Simulator Screenshot - iPhone 17 Pro Max - 2026-07-27 at 18 26 08
Simulator Screenshot - iPhone 17 Pro Max - 2026-07-27 at 18 41 26

blotztask-api/Modules/Referrals/Commands/RedeemReferralCode.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public async Task Handle(RedeemReferralCodeCommand command, CancellationToken ct = default)
+    {
+        var code = command.Code.Trim().ToUpperInvariant();
+
+        var referee = await db.AppUsers.FirstOrDefaultAsync(u => u.Id == command.RefereeUserId, ct)
+            ?? throw new NotFoundException("User not found.");
+
+        if (referee.IsOnboarded)
+            throw new InvalidOperationException("Invite codes can only be redeemed during onboarding.");
```
sol-wizard: **Major — no existing user can ever redeem a code.**

This gate, combined with the only redeem UI being `OnboardingInviteSection` (the Settings screen is view/copy only), means every already-onboarded user can share a code but never redeem one. Tapping Skip during onboarding also loses the chance permanently — there's no recovery path.

Is the one-shot design intentional? If so, the release note should say redemption happens at signup; "find it in Settings" reads as two-way.

blotztask-api/Modules/Referrals/Commands/RedeemReferralCode.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        var referralCode = await db.ReferralCodes.FirstOrDefaultAsync(r => r.Code == code, ct)
+            ?? throw new NotFoundException("Referral code not found.");
+
+        if (referralCode.OwnerUserId == command.RefereeUserId)
+            throw new ArgumentException("You cannot redeem your own referral code.");
+
+        var alreadyRedeemed = await db.Referrals.AnyAsync(r => r.RefereeUserId == command.RefereeUserId, ct);
+        if (alreadyRedeemed)
+            throw new InvalidOperationException("You have already redeemed a referral code.");
```
sol-wizard, line 36: **Major — the user sees the wrong error message.**

This and line 26 both throw `InvalidOperationException`, which `ErrorHandlerMiddleware` maps to **409**. The app then shows *"You have already redeemed an invite code"* for both cases — so someone blocked by the onboarding rule is told they already redeemed.

Suggest distinct exception types, or a machine-readable error code in `ApiResponse` that the client switches on.

blotztask-mobile/src/shared/util/queryClient.ts
```diff
@@ -70,11 +70,16 @@ export const queryClient = new QueryClient({
...
-function getErrorMessage(error: unknown): string {
+function getErrorMessage(error: unknown, meta?: Record<string, unknown>): string {
   if (isAxiosError(error)) {
     const status = error.response?.status;
     if (!error.response) return i18n.t("errors.network");
     if (status && status >= 500) return i18n.t("errors.server");
+    if (meta?.errorNs === "invite") {
+      if (status === 404) return i18n.t("settings:invite.errors.notFound");
+      if (status === 409) return i18n.t("settings:invite.errors.alreadyRedeemed");
+      return i18n.t("settings:invite.errors.invalid");
```
sol-wizard: **Major** — this fallback returns *"You cannot use your own invite code"* for **any** non-404/409 status, including 401 and 403. Only 400 actually means self-referral.

Switching on a server-supplied error code rather than HTTP status would fix this and the 409 collision noted in `RedeemReferralCode.cs`.

blotztask-mobile/src/shared/util/queryClient.ts
```diff
@@ -70,11 +70,16 @@ export const queryClient = new QueryClient({
   },
 });
 
-function getErrorMessage(error: unknown): string {
+function getErrorMessage(error: unknown, meta?: Record<string, unknown>): string {
   if (isAxiosError(error)) {
     const status = error.response?.status;
     if (!error.response) return i18n.t("errors.network");
     if (status && status >= 500) return i18n.t("errors.server");
+    if (meta?.errorNs === "invite") {
```
sol-wizard: **Minor** — feature-specific logic in a shared util. Each new feature adds another branch here. A generic `meta.errorMap` passed from the mutation would keep this function neutral.

blotztask-api/Infrastructure/Data/Configurations/ReferralConfiguration.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace BlotzTask.Infrastructure.Data.Configurations;
+
+public class ReferralConfiguration : IEntityTypeConfiguration<Referral>
+{
+    public void Configure(EntityTypeBuilder<Referral> builder)
+    {
+        builder.ToTable("Referrals");
+        builder.HasKey(r => r.Id);
+        builder.Property(r => r.CodeUsed).IsRequired().HasMaxLength(12);
+        builder.HasIndex(r => r.RefereeUserId).IsUnique();
```
sol-wizard, line 15: **Minor — `Referrals` has no foreign keys.**

No `HasOne` for `ReferrerUserId` or `RefereeUserId`, and the migration confirms zero FKs. `ReferralCodes` does cascade. Given the existing `AddCascadeDeletesForUserRelations` migration, deleting a user will now leave orphan referral rows.

sol-wizard (commented):
> 🤖 **AI-generated PR review.** Treat the inline comments as suggestions, not verdicts — use your own judgement and verify each before changing anything.

Core flow looks sound. The three `Major` inline comments are what I'd resolve before merge.

**One PR-level issue (no line to anchor to): the description doesn't match the code.**
It describes an `InviteRedemptions` table, an `InviteCode` column on `AppUsers`, `/api/invite/*` routes, and a redeem input in Settings. Actual: `ReferralCodes` + `Referrals`, `api/referral/*`, and no redeem input in Settings — the migration *drops* `InviteRedemptions` and `InviteCode`. Looks like the description describes the first iteration. Worth updating since `pr-release-note-check.yml` gates on it.

<details><summary>Nits — feel free to ignore</summary>

- `nit:` naming split across the stack — mobile `feature/invite/` + `InviteCodeDTO` vs backend `Referrals`/`ReferralCode`/`api/referral`.
- `nit:` whitespace-only edits in `AppUserConfiguration.cs`, `AppUser.cs`, `BlotzTaskDbContext.cs`.
- `speculative — feel free to ignore:` a concurrent first-time `GET /my-code` could double-insert and surface a 409 from the unique index; retry fixes it, and the double-redeem race is correctly absorbed by `DbUpdateException` → 409. I'd leave both.

</details>

**Test worth adding:** one handler test over `RedeemReferralCode` covering self-referral / already-redeemed / onboarding-gate. User-isolation logic, cheap with an in-memory context, and the comment on line 36 shows those branches are easy to conflate.

**Assumptions:** diff-only review, nothing built or run. I assumed `InviteRedemptions` / `AppUsers.InviteCode` were merged to main earlier, since this migration drops them without creating them — **please confirm they were never populated in staging/prod**, as `DropTable` is destructive with no data-migration step. I did not check badge seed config for the new enum values.

blotztask-api/Modules/Users/Commands/SyncUser.cs
```diff
@@ -139,12 +141,16 @@ public async Task<SyncUserResult> Handle(SyncUserCommand command, CancellationTo
...
 
+            await ensureReferralCode.HandleAsync(row.Id, ct);
+
             return new SyncUserResult { Id = row.Id, Auth0UserId = row.Auth0UserId };
         }
 
         existing.LoginAt = utcNow;
         await db.SaveChangesAsync(ct);
 
+        await ensureReferralCode.HandleAsync(existing.Id, ct);
```
sol-wizard: **Minor — extra query on every login.**

`GetMyReferralCode` already lazy-creates via `EnsureReferralCode`, so this call on the existing-user path adds a redundant SELECT to one of the hottest endpoints. The new-user call on line 144 should be enough.

blotztask-api/Modules/Referrals/Services/ReferralCodeGenerator.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    private const string Alphabet = "ABCDEFGHJKLMNPQRSTUVWXYZ23456789";
+
+    private readonly SqidsEncoder<int> _sqids = new(new SqidsOptions
+    {
+        Alphabet = Alphabet,
+        MinLength = 8
+    });
+
+    public string Encode(int id) => _sqids.Encode(id);
```
sol-wizard, line 15: **Minor — invite codes are enumerable.**

This encodes the sequential `ReferralCode.Id` with a fixed public alphabet and no salt. Sqids is explicitly not an obfuscation mechanism, so valid codes can be derived rather than guessed.

Harmless today since redeeming grants nothing — but `TriggerAction.InviteRedeemed` / `EventValueKey.InviteCount` suggest a reward is coming, and then this becomes farmable. A random code with a uniqueness retry would close it before that lands.

blotztask-api/Modules/Badges/Enum/TriggerAction.cs
```diff
@@ -2,5 +2,6 @@ namespace BlotzTask.Modules.Badges.Enum;
 
 public enum TriggerAction
 {
-    TaskComplete
+    TaskComplete,
+    InviteRedeemed
```
sol-wizard, line 7: **Minor — dead value.** Nothing publishes this event; `RedeemReferralCode` emits no badge event. Same for `EventValueKey.InviteCount`. Worth wiring up or dropping until the badge actually lands.

blotztask-api/Infrastructure/Data/Migrations/20260727164219_IncreaseCodeUsedMaxLength.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace BlotzTask.Infrastructure.Data.Migrations
+{
+    /// <inheritdoc />
+    public partial class IncreaseCodeUsedMaxLength : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AlterColumn<string>(
```
sol-wizard: **Minor — squash before merge.** This only widens a column that `MakeReferralCodeNullable` created minutes earlier.

Also that migration's name is misleading: it drops a table, drops a column, and creates two new ones.

blotztask-api/Infrastructure/Data/Migrations/20260727150250_ReplaceInviteSystemWithReferralSystem.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropTable(
+                name: "InviteRedemptions");
+
+            migrationBuilder.DropColumn(
+                name: "InviteCode",
+                table: "AppUsers");
```
lyx2-2: Migration will fail against main
20260727150250_ReplaceInviteSystemWithReferralSystem Up drops InviteRedemptions and AppUsers.InviteCode, but those objects are not on main (only from this branch’s earlier invite iteration). Fresh envs applying this after merge will fail.

blotztask-mobile/src/feature/invite/hooks/useRedeemInviteCode.ts
```diff
@@ -0,0 +1,14 @@
+import { useMutation } from "@tanstack/react-query";
+import { redeemInviteCode } from "../services/invite-service";
+
+export const useRedeemInviteCode = () => {
+  const mutation = useMutation({
+    mutationFn: (code: string) => redeemInviteCode(code),
+    meta: { errorNs: "invite" },
```
lyx2-2: errorNs isn’t used by queryClient, it only reads meta.errorMap. So the invite error strings in settings.json never show. Could pass an errorMap keyed by status code instead.

lyx2-2 (dismissed):

blotztask-mobile/src/feature/invite/hooks/useRedeemInviteCode.ts
LiChenyang-GZ: **The localised invite error messages never fire — the "Localised error messages per error type" claim in the description isn't true yet.**

This sets `meta.errorNs`, but the global handler reads `meta.errorMap` (`queryClient.ts:78`). `errorNs` isn't consumed anywhere in the repo, so the new `invite.errors.*` keys in `settings.json` are dead code and users get the generic `errors.default` / `errors.validation`.

Worse: **403 never reaches that handler at all.** `mutationCache.onError` early-returns on 401 *and* 403 (`queryClient.ts:31-35`), so redeeming after onboarding produces no toast and no Sentry event — literally nothing happens on screen.

Suggested fix:

1. Narrow the early return to 401 only. 401 is the token-refresh path that legitimately needs to be silent; 403 means "you can't do this" and should surface. Safe to change — the `ForbiddenException` added in this PR is the only 403 source in the entire API.
2. Move the copy into `common.json` under `errors.invite.*` (en + zh), where the other `getErrorMessage` strings already live, and delete the dead `settings.json` keys. Note 403 has no copy at all right now.
3. Then here:

```ts
meta: {
  errorMap: {
    400: "errors.invite.selfCode",
    403: "errors.invite.onboardingOnly",
    404: "errors.invite.notFound",
    409: "errors.invite.alreadyRedeemed",
  },
},
```

Root cause worth fixing separately: `meta` is untyped (`Record<string, unknown>`), which is exactly why `errorNs` compiled without complaint. A `declare module "@tanstack/react-query" { interface Register { mutationMeta: ... } }` augmentation would catch this class of typo at build time. (It has to be declared as a `type`, not an `interface` — TanStack silently falls back to the default untyped meta if it isn't assignable to `Record<string, unknown>`.)

blotztask-api/Modules/Referrals/Commands/RedeemReferralCode.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public async Task Handle(RedeemReferralCodeCommand command, CancellationToken ct = default)
+    {
+        var code = command.Code.Trim().ToUpperInvariant();
+
+        var referee = await db.AppUsers.FirstOrDefaultAsync(u => u.Id == command.RefereeUserId, ct)
+            ?? throw new NotFoundException("User not found.");
+
+        if (referee.IsOnboarded)
+            throw new ForbiddenException("Invite codes can only be redeemed during onboarding.");
```
LiChenyang-GZ, line 26: **Race with the onboarding flow — this guard can reject a legitimate redeem.**

On the client `redeemInviteCode` is fire-and-forget (`mutation.mutate`, not awaited), while `onboarding-screen.tsx`'s `handleFinish` calls `setUserOnboarded(true)`. If the user taps Redeem and then immediately taps Get Started, the two requests race; if the profile PATCH lands first, this guard rejects the redeem with 403. Combined with the `errorMap` issue on the mobile side, the user sees nothing at all and just silently loses the referral.

Minimal fix: disable the footer button while the redeem is in flight. The button and the mutation live in different components, so the tidiest way is to give the mutation a `mutationKey` and read `useIsMutating({ mutationKey })` in `onboarding-screen.tsx`.

blotztask-api/Modules/Referrals/Queries/GetMyReferralCode.cs
```diff
@@ -0,0 +1,24 @@
...
+}
+
+public class GetMyReferralCodeQueryHandler(BlotzTaskDbContext db, EnsureReferralCodeHandler ensureReferralCode)
+{
+    public async Task<ReferralCodeDto> Handle(GetMyReferralCodeQuery query, CancellationToken ct = default)
+    {
+        await ensureReferralCode.HandleAsync(query.UserId, ct);
+
+        var referralCode = await db.ReferralCodes
+            .FirstAsync(r => r.OwnerUserId == query.UserId, ct);
```
LiChenyang-GZ: `FirstAsync` throws `InvalidOperationException("Sequence contains no elements")` when there's no row, and `ErrorHandlerMiddleware` maps `InvalidOperationException` to **409 Conflict**, returning that raw message to the client. Wrong status plus an internal string leak.

The cleanest fix isn't `FirstOrDefaultAsync` — it's to have `EnsureReferralCodeHandler.HandleAsync` return the `ReferralCode` it already has in hand, which removes this second round-trip entirely:

```csharp
public async Task<ReferralCode> HandleAsync(Guid userId, CancellationToken ct = default)
```

Then this handler becomes a single line and the `Code!` null-forgiving operator goes away too.

blotztask-api/Modules/Referrals/Services/ReferralCodeGenerator.cs
LiChenyang-GZ, line 15: *speculative — feel free to ignore for now:* codes are Sqids over the identity `Id`, so they're sequential and enumerable (the Sqids docs are explicit that it is not a security primitive). Harmless today because redeeming grants nothing, but `EventValueKey.InviteCount` / `TriggerAction.InviteRedeemed` suggest rewards are on the way — at that point someone can farm referrals by walking the code space. Worth adding a random component before any reward ships.

LiChenyang-GZ (commented):
> **AI-generated review.** These are suggestions from an automated senior-level pass, not verdicts. Please verify each point yourself before acting — some may be wrong or missing context I don't have.

Solid feature overall. Things that are clearly right: the unique index on `Referrals.RefereeUserId` plus the `Restrict` FKs means a concurrent double-redeem gets caught by the DB even if it slips past the `AnyAsync` check (and `DbUpdateException` maps to 409, which happens to be the correct status). `DeleteUser` clearing `Referrals` manually while `ReferralCodes` cascades is the right split. `ToUpperInvariant()` normalisation lines up with the uppercase-only generator alphabet, and dropping I/O/0/1 from it is a nice touch.

`dotnet build` passes locally (0 errors).

Inline comments cover the concrete code issues. Three things that don't fit inline:

### 1. The PR description doesn't match the diff

| Description says | Actual |
|---|---|
| `InviteRedemptions` table + `AppUsers.InviteCode` column | `ReferralCodes` + `Referrals` tables |
| `GET/POST /api/invite/*` | `/api/referral/my-code`, `/api/referral/redeem` |
| Settings has an input to redeem another user's code | Settings only shows/copies your own code; redeem lives in onboarding |
| — | No mention of the 403 "onboarding only" restriction, which is the single biggest behavioural constraint in this PR |
| — | No mention that onboarding goes from 3 steps to 4 |

This affects the release note too: it says "find it in Settings", and users can indeed see their code there, but there is no entry point anywhere to *use* someone else's code after onboarding. As written the checklist is misleading.

### 2. Question: is the onboarding-only window intended?

`handleFinish` in `onboarding-screen.tsx` is shared by both the Skip button and Get Started, so either path sets `IsOnboarded = true` permanently. Two consequences: a user who taps Skip can never redeem, and every existing user is already onboarded, so none of them can participate at all.

If that's a deliberate anti-abuse decision it's fine — just worth stating in the description. If not, either add a redeem entry point in Settings, or relax the backend rule to something like "within N days of signup and not yet redeemed".

### 3. Minor / nits — feel free to ignore

- `EnsureReferralCode`'s two-phase write (insert with `Code = null`, then update) exists only because the code is derived from the identity `Id`. The cost is that `Code` stays nullable forever and every new user pays two `SaveChanges`. A random 8-char code from the same alphabet plus a uniqueness retry would let `Code` be `IsRequired()`. Both designs are defensible; the current one works.
- `EventValueKey.InviteCount` / `TriggerAction.InviteRedeemed` are unused `//To do` placeholders — probably cleaner to add them in the badge PR that actually consumes them.
- Naming is split three ways: backend `Referral`, route `/api/referral`, mobile `feature/invite` and "invite" copy. Settling on one word would save real grep time later.
- Whitespace-only edits in `AppUser.cs`, `AppUserConfiguration.cs` and `BlotzTaskDbContext.cs` add diff noise.

### One test worth writing

`RedeemReferralCodeCommandHandler`'s three guards (self-referral, double redeem, onboarding gate) are user-isolation logic that breaks silently when the onboarding rule changes, and they're short to cover with an in-memory/SQLite context. I wouldn't add anything beyond that here.

blotztask-api/Modules/Badges/Enum/EventValueKey.cs
```diff
@@ -2,5 +2,7 @@ namespace BlotzTask.Modules.Badges.Enum;
 
 public enum EventValueKey
 {
-    CompleteOffsetMins
+    CompleteOffsetMins,
+    //To do
```
sol-wizard, line 6: what is this todo ? 

example of TODO: //TODO: Need to complete this section and use in reward once product manager confirm what to do with the invitation badge

sol-wizard (approved):

PR #478, opened 2026-07-19
fix widget

LiChenyang-GZ:
## Summary

solve the crash in app by fixing the widget part

## Release note

> none

**Status:**


blotztask-mobile/src/feature/widget/ios/services/sync-ios-today-tasks-widget-cache.ts
```diff
@@ -14,7 +15,8 @@ export async function syncIosTodayTasksWidgetCache(
           props: snapshot,
         })),
     );
-  } catch {
-    // Widget timeline updates are best-effort.
```
sol-wizard, line 18: if you catch you need to do something , if not you swallow the error

sol-wizard (changes requested):

sol-wizard (approved):

PR #479, opened 2026-07-19
Add What's New onboarding skill and simulator build profile

sol-wizard:
## Summary

Adds the **"What's New" update-onboarding page** — a swipeable carousel shown once per release to onboarded users (gated on a per-version `AsyncStorage` key). Also extracts a shared `IntroCarousel` now used by both onboarding and What's New, plus en/zh copy.

Ships alongside the Claude tooling to produce its content monthly: the `generate-whatsnew` skill and a `preview-simulator` EAS profile for simulator screenshots. That screenshot flow hasn't been run end to end yet — first runner should note breakage in `SKILL.md`.

## Release note

> New "What's New" page highlights the latest features each release.

**Status:**


.claude/skills/generate-whatsnew/blotz-whatsnew.html
sol-wizard, line 1: do we need this html to be in our repo ?

blotztask-mobile/src/app/(protected)/whats-new.tsx
```diff
@@ -0,0 +1,5 @@
+import WhatsNewScreen from "@/feature/whats-new/screens/whats-new-screen";
+
+export default function WhatsNewRoute() {
```
sol-wizard: **This route is never reached — dead code.** The gate renders `<WhatsNewScreen />` inline in `index.tsx`, and nothing in the app navigates to `/(protected)/whats-new` (grepped `src/app` + `src/feature` for `router.*` / `href` / `Link` → zero hits). Pick one approach: either `router.replace("/(protected)/whats-new")` from the gate and drop the inline render, **or** delete this file and keep the inline render. Keeping both leaves an orphaned route.

blotztask-mobile/src/app/(protected)/index.tsx
```diff
@@ -27,15 +30,26 @@ export default function ProtectedGate() {
...
   }, [
     isUserProfileLoading,
     isUserPreferencesLoading,
     isUpdatingUserPreferences,
     userProfile?.isOnBoarded,
     userPreferences?.preferredLanguage,
+    hasSeenWhatsNew,
   ]);
 
+  if (userProfile?.isOnBoarded && hasSeenWhatsNew === false) {
```
sol-wizard, line 55: **Confirm intent: new users will hit this on their 2nd launch.** Onboarding finishes with a direct `router.replace("/(protected)/(tabs)")`, so a brand-new user doesn't see What's New immediately — but on their next cold start they land here with `isOnBoarded === true` and `hasSeenWhatsNew === false`, so they get a "What's New" tour for features they never knew the app without. If that's undesired, call `markAsSeen()` when onboarding completes.

blotztask-mobile/src/feature/whats-new/screens/whats-new-screen.tsx
```diff
@@ -0,0 +1,155 @@
...
+      </Text>
+      <Text className="text-base font-baloo text-black/40 text-center max-w-xs">
+        {t(card.bodyKey)}
+      </Text>
+    </View>
+  );
+}
+
+export default function WhatsNewScreen() {
+  const { markAsSeen } = useWhatsNewSeen();
```
sol-wizard: **Separate hook instance from the one in `index.tsx` — minor/fragile.** The gate reads `hasSeen` from its own `useWhatsNewSeen()`; this one owns `markAsSeen`. They don't share state, so it only works because `handleFinish` navigates away before the gate's stale `hasSeen` matters. Fine today, but a footgun if this screen is ever rendered without an immediate `router.replace`. Passing `markAsSeen` down (or a shared context) would be more robust.

blotztask-mobile/src/feature/whats-new/screens/whats-new-screen.tsx
```diff
@@ -0,0 +1,155 @@
...
+          </View>
+        </GradientColor>
+        <View className="flex-1 items-end">
+          <Pressable onPress={handleFinish} hitSlop={10}>
+            <Text className="text-xl font-baloo text-black/40">{t("actions.skip")}</Text>
+          </Pressable>
+        </View>
+      </View>
+
+      <FlatList
```
sol-wizard: _Optional / your call:_ this `FlatList` + paging + dot-indicator + `onMomentumScrollEnd` block is nearly identical to `onboarding-screen.tsx`. Not a blocker — could extract a shared `<Carousel>` now or leave it for later.

sol-wizard (commented):
A few inline notes — line by line below. Overall solid; the versioned AsyncStorage gate is a nice design. Worth running `expo lint` before merge since none of this has been exercised yet.

blotztask-mobile/src/shared/components/intro-carousel.tsx
```diff
@@ -0,0 +1,126 @@
...
+        pagingEnabled
+        showsHorizontalScrollIndicator={false}
+        getItemLayout={(_, index) => ({
+          length: SCREEN_WIDTH,
+          offset: SCREEN_WIDTH * index,
+          index,
+        })}
+        onMomentumScrollEnd={onMomentumScrollEnd}
+        renderItem={({ item }) => (
+          <View style={{ width: SCREEN_WIDTH, flex: 1 }}>{renderItem(item)}</View>
```
sol-wizard, line 99: nit (verify visually, feel free to ignore): the old onboarding page wrapper was `<View style={{ width: SCREEN_WIDTH }}>` with no `flex: 1` — this shared version adds it, and the header also dropped `pt-2` and bumped the Blotz text `leading-9`→`leading-10`. What's New needs `flex: 1` (its screenshot card fills height), but onboarding didn't have it. Worth a quick side-by-side on the simulator to confirm onboarding still lays out the same; if it shifts, we can make these per-screen props.

sol-wizard (commented):
Re-review — all four earlier comments resolved (dead route deleted, onboarding now marks What's New seen on finish, single markAsSeen source via onFinish prop, shared IntroCarousel extracted). Nice work going the full shared-component route. One cosmetic nit below — AI-assisted, please verify before acting.

lyx2-2 (approved):

PR #480, opened 2026-07-19
Add one-time Apple Calendar export

dksjl661:
## Summary

- add an iOS 17+ action that opens a prefilled Apple Calendar event form from task details
- request write-only calendar access and keep exported events independent from Blotz tasks
- export only the selected occurrence of a recurring task as a one-time event
- update English and Chinese UI and privacy copy

## Validation

- `npm install`
- `npx eslint . --ext .ts,.tsx,.js`
- verified the resolved Expo config contains only the iOS write-only calendar usage key and removes Android calendar permissions

## Release note

> Add a Blotz task to Apple Calendar as a one-time, editable event.

**Status:**


dksjl661:
image
image

blotztask-mobile/src/feature/task-details/screens/task-details-screen.tsx
```diff
@@ -145,6 +162,57 @@ export default function TaskDetailsScreen() {
     });
   };
 
+  const handleExportToAppleCalendar = async () => {
+    if (!selectedTask || isExportingToCalendar) return;
+
+    setIsExportingToCalendar(true);
+
+    try {
+      const permission = await requestCalendarPermissions(true);
```
Guotai812: it seems like the calendar access doesnt need to get permission, try whether we can add to clendar without permission

Guotai812 (approved):

PR #482, opened 2026-07-21
Delete function for ai voice cards

LiliZ3681:
## Summary

Updated the delete behaviour for AI-generated cards. Swiping past the delete threshold now deletes the card automatically, while the delete button remains available if the threshold is not reached.
## Release note

> _Swipe past the delete threshold to delete AI-generated cards._

**Status:**

https://github.com/user-attachments/assets/89fb16e9-5298-4979-bf5d-8cc30fd58683


dksjl661 (approved):
Terrific!!!

PR #485, opened 2026-07-24
Track task completion, reopen, and delete in PostHog

LiChenyang-GZ:
## Summary

Adds two PostHog events — `task_completed`, `task_reopened`, `task_deleted` — for
normal and recurring tasks. We currently track task creation but not
completion; this adds the completion side so we can measure whether and
how users finish their tasks, and later connect creation → completion by
`task_id`. Additive only — no existing events removed, no backend/DB
change. Events fire on server-confirmed success and only on genuine
completion (same guard as the completion firework).

## Release note

> none

**Status:**

### DIFF
--- a/blotztask-mobile/src/feature/calendar/components/task-card.tsx
+++ b/blotztask-mobile/src/feature/calendar/components/task-card.tsx
@@ -307,11 +307,19 @@ const TaskCard = ({ task, deleteTask, isDeleting, selectedDay, onOpenMode, onRow
                       recurringTaskId,
                       occurrenceDate,
                       wasDone: task.isDone,
+                      wasOverdue: isOverdue,
+                      hasDeadline: task.isDeadline,
                     });
                     return;
                   }
                   if (!hasTaskItemId(task)) return;
-                  toggleTask({ taskId: task.id, selectedDay, wasDone: task.isDone });
+                  toggleTask({
+                    taskId: task.id,
+                    selectedDay,
+                    wasDone: task.isDone,
+                    wasOverdue: isOverdue,
+                    hasDeadline: task.isDeadline,
+                  });
                   if (task.alertTime && new Date(task.alertTime) > new Date()) {
                     await cancelNotification({ notificationId: task?.notificationId });
                   }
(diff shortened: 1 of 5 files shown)


LiChenyang-GZ:
## Details

### 1. How the tracking works
- `task_completed` / `task_reopened` defined in `posthog-events.ts`; capture methods in `analytics.ts`.
- Fired in the mutation `onSuccess` (server-confirmed), not on the UI tap → no optimistic double-counting on rollback.
- `wasDone` (state before the tap) decides completed vs reopened — same guard as the `playIfCompleting` firework, so only real completions count.
- Covers both paths: normal tasks (`useTaskMutations`) and recurring occurrences (`useRecurringTaskMutations`).
- Properties: `task_id` (join key), `is_recurring`, `was_overdue`, `has_deadline`, `occurrence_date` (recurring only).
- `task_completed` / `task_reopened` / `task_deleted` defined in posthog-events.ts; capture methods in analytics.ts.
- Deletion fires in `deleteTaskMutation` / `deleteRecurringOccurrenceMutation onSuccess`; no `wasDone` guard needed (delete is unambiguous). Normal deletes carry `was_overdue` / `has_deadline`; recurring deletes carry `occurrence_date`.


### 2. Why
- We track task creation but not completion — the value moment of a to-do app. This lets us measure completion behavior.
- `task_id` is a join key: once creation is also tracked with `task_id` (next PR), created → completed can be joined. Preset tasks are seeded server-side and never fire a creation event, so this naturally separates self-created from preset tasks — no DB change.

### 3. What we can learn
- With the existing `active_user_5s`: whether completing tasks relates to users returning (Retention insight); completion volume over time; the share of completions that were overdue; deadline vs normal; recurring vs one-off.
- Note: until the creation event lands, raw completion counts include completions of the preset starter tasks; cohorts are small, so read weekly rather than daily.
- Deletion rate over time; and once creation is tracked, whether AI-created tasks get deleted more than manual ones.

### 4. What each finding could suggest → next step
- Completers return notably more often → "complete a task" is likely the activation moment → help new users reach their first completion sooner → build the activation funnel (needs the creation event, next PR).
- Share of overdue completions trending up → tasks are piling up before they get done → look at reminders / scheduling.
- Completion shows little relation to returning → completion isn't the main driver → look elsewhere (e.g. AI reliability, onboarding).
- High delete rate on AI-created tasks → AI output quality problem → look at the generation prompt / acceptance flow.

dksjl661 (approved):

PR #496, opened 2026-07-29
Fix badge detail sharing and achievement title alignment

FFFLAREEE:
## Summary

- Enable system image sharing from the badge detail page.
- Modify badge popout UI.
- Adjust English and Chinese achievement titles for improved visual centering.

## Release note

> Badge detail pages can now share badge images, with improved achievement-title alignment.

**Status:**

### DIFF
--- a/blotztask-mobile/src/feature/badge/components/badge-share-card.tsx
+++ b/blotztask-mobile/src/feature/badge/components/badge-share-card.tsx
@@ -4,15 +4,13 @@ import { forwardRef } from "react";
 import { useTranslation } from "react-i18next";
 import { Image, Text, View } from "react-native";
 
-interface BadgeShareCardBadge {
-  name: string;
-  description: string;
-  iconUrl: string;
-  obtainedAt?: Date;
-}
-
 interface BadgeShareCardProps {
-  badge: BadgeShareCardBadge;
+  badge: {
+    name: string;
+    description: string;
+    iconUrl: string;
+    obtainedAt: Date;
+  };
 }
 
 function ShareCardBackground() {
@@ -63,13 +61,11 @@ export const BadgeShareCard = forwardRef<View, BadgeShareCardProps>(function Bad
             {badge.description}
           </Text>
 
-          {badge.obtainedAt ? (
-            <Text className="mt-4 scale-110 text-secondary/40 text-sm font-baloo text-center">
-              {t("obtainedOn", {
-                date: formatLocalizedDate(badge.obtainedAt, "fullMonthDayYear"),
-              })}
-            </Text>
-          ) : null}
+          <Text className="mt-4 scale-110 text-secondary/40 text-sm font-baloo text-center">
+            {t("obtainedOn", {
+              date: formatLocalizedDate(badge.obtainedAt, "fullMonthDayYear"),
+            })}
+          </Text>
         </View>
 
         <View
--- a/blotztask-mobile/src/feature/badge/components/badge-share-card.tsx
+++ b/blotztask-mobile/src/feature/badge/components/badge-share-card.tsx
@@ -1,12 +1,18 @@
-import { BadgeNotificationDTO } from "@/feature/badge/models/badge-notification-dto";
 import { ASSETS } from "@/shared/constants/assets";
 import { formatLocalizedDate } from "@/shared/util/localized-date-format";
 import { forwardRef } from "react";
 import { useTranslation } from "react-i18next";
 import { Image, Text, View } from "react-native";
 
+interface BadgeShareCardBadge {
+  name: string;
+  description: string;
+  iconUrl: string;
+  obtainedAt?: Date;
+}
+
 interface BadgeShareCardProps {
-  badge: BadgeNotificationDTO;
+  badge: BadgeShareCardBadge;
 }
 
 function ShareCardBackground() {
@@ -49,9 +55,7 @@ export const BadgeShareCard = forwardRef<View, BadgeShareCardProps>(function Bad
             {t("shareCardHeadline")}
           </Text>
 
-          <Text
-            className="mt-4 w-full text-secondary text-[25px] font-balooBold text-center leading-9"
-          >
+          <Text className="mt-4 w-full text-secondary text-[25px] font-balooBold text-center leading-9">
             {badge.name}
           </Text>
 
@@ -59,11 +63,13 @@ export const BadgeShareCard = forwardRef<View, BadgeShareCardProps>(function Bad
             {badge.description}
           </Text>
 
-          <Text className="mt-4 scale-110 text-secondary/40 text-sm font-baloo text-center">
-            {t("obtainedOn", {
-              date: formatLocalizedDate(badge.obtainedAt, "fullMonthDayYear"),
-            })}
-          </Text>
+          {badge.obtainedAt ? (
+            <Text className="mt-4 scale-110 text-secondary/40 text-sm font-baloo text-center">
+              {t("obtainedOn", {
+                date: formatLocalizedDate(badge.obtainedAt, "fullMonthDayYear"),
+              })}
+            </Text>
+          ) : null}
         </View>
 
         <View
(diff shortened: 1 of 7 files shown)


FFFLAREEE:
216f5a38e60fa3bfc5a1077dbe91d4e7
143aca5a7a291ae2724bb32eb8a6d829
272a5b0a5e90305b809abd3dddce0172
c4c73d8423341d12a08e373dadbba7e0

blotztask-mobile/src/feature/badge/components/badge-share-card.tsx
```diff
@@ -1,12 +1,18 @@
...
 import { formatLocalizedDate } from "@/shared/util/localized-date-format";
 import { forwardRef } from "react";
 import { useTranslation } from "react-i18next";
 import { Image, Text, View } from "react-native";
 
+interface BadgeShareCardBadge {
+  name: string;
+  description: string;
+  iconUrl: string;
+  obtainedAt?: Date;
```
sol-wizard: **major:** on the badge-detail path this is always `undefined` — `BadgeDetailDTO` has no `obtainedAt`, and `GetBadgeById.cs` never returns one. So the date isn't "omitted when unavailable", it's dropped every time: the same badge shared from the achievement popup shows "Obtained on …", from the detail page it shows nothing.

The data exists — `UserBadge.EarnedAtUtc`, and `GetBadgeByIdQuery` already takes `UserId` without using it. Suggest joining `UserBadges` and adding `ObtainedAt` to `BadgeByIdItemDto` + `BadgeDetailDTO`.

If shipping without the date is deliberate, worth rewording the PR description — the current phrasing reads as an edge-case guard.

blotztask-mobile/src/feature/badge/components/badge-share-card.tsx
```diff
@@ -49,21 +55,21 @@ export const BadgeShareCard = forwardRef<View, BadgeShareCardProps>(function Bad
...
           <Text className="mt-3 w-full text-secondary/60 text-[15px] font-baloo text-center leading-[18px]">
             {badge.description}
           </Text>
 
-          <Text className="mt-4 scale-110 text-secondary/40 text-sm font-baloo text-center">
-            {t("obtainedOn", {
-              date: formatLocalizedDate(badge.obtainedAt, "fullMonthDayYear"),
-            })}
-          </Text>
+          {badge.obtainedAt ? (
```
sol-wizard: _speculative — feel free to ignore:_ `GetBadgeById` accepts `UserId` but never checks ownership, so any badge ID is fetchable. Harmless today since the wall lists only earned badges — but this card claims *"I've unlocked:"*, which becomes a false claim if the wall ever shows locked badges. The `UserBadges` join suggested above would cover it.

blotztask-mobile/src/i18n/locales/en/badge.json
```diff
@@ -1,5 +1,5 @@
 {
-  "achievementUnlocked": "Achievement unlocked!",
+  "achievementUnlocked": " Achievement unlocked!",
```
sol-wizard, line 2: **major:** a leading space is fragile as a centering fix — invisible in review and in diffs, and translation tooling routinely trims leading whitespace, so it can silently revert.

It also isn't scoped to the screen being fixed: `achievementUnlocked` renders in both `badge-details-screen.tsx:54` and `badge-achievement-modal.tsx:65`. The modal constrains it with `max-w-64` so the text wraps — the space shifts line 1 only, a different result on a surface this PR didn't screenshot.

Suggest reverting both locales and compensating at the call site instead: `paddingLeft` on the `<Text>`, or inside `GradientColor` if the MaskedView is what's offsetting it.

blotztask-mobile/src/i18n/locales/zh/badge.json
```diff
@@ -1,5 +1,5 @@
 {
-  "achievementUnlocked": "成就解锁！",
+  "achievementUnlocked": " 成就解锁！",
```
sol-wizard, line 2: Same issue as the `en` locale — worth reverting alongside it.

blotztask-mobile/src/feature/badge/screens/badge-details-screen.tsx
```diff
@@ -30,59 +38,67 @@ export default function BadgeDetailsScreen() {
...
-              className="h-14 min-w-36 rounded-full bg-highlight px-5 items-center justify-center shadow-lg shadow-lime-300"
-              onPress={() => console.log("Share reward", badgeDetail.id)}
-            >
-              <Text className="text-lg font-balooBold text-white">{t("shareReward")}</Text>
-            </Pressable>
-          </View>
+            <View className="mt-8 flex-row items-center justify-center gap-4">
+              <Pressable
+                className="h-14 min-w-36 rounded-full border border-highlight bg-white/60 px-5 items-center justify-center"
+                onPress={() => console.log("Equip reward", badgeDetail.id)}
```
sol-wizard, line 85: **minor** (pre-existing, not introduced here): Equip is still a `console.log` while Share now works right beside it, and this PR is tagged **User-facing — announce it**. Worth confirming the release note won't point users at a button that does nothing.

blotztask-mobile/src/feature/badge/screens/badge-details-screen.tsx
```diff
@@ -7,9 +7,17 @@ import { useTranslation } from "react-i18next";
 import { Pressable, ScrollView, Text, View } from "react-native";
 import { SafeAreaView } from "react-native-safe-area-context";
 import { useBadgeDetailQuery } from "../hooks/useBadgeDetailQuery";
+import { BadgeShareCard } from "@/feature/badge/components/badge-share-card";
+import { useReviewShare } from "@/feature/review/hooks/useReviewShare";
```
sol-wizard, line 11: _nit — feel free to ignore:_ `useReviewShare` now has two badge consumers and toasts `settings:review.shareError`. Worth promoting to something like `shared/hooks/useShareImage`.

sol-wizard (commented):
## AI-generated PR review

_Generated by an AI assistant. Treat the findings as suggestions — use your own judgement and verify each one before changing anything._

CI is green (8/8 checks, 98 tests); branch is `behind` main so it'll need a rebase. The core change looks sound — reusing the `useReviewShare` + off-screen `BadgeShareCard` pattern from the achievement modal is the right approach, and narrowing the card's prop type off `BadgeNotificationDTO` is a good call.

2 major, 1 minor and 2 nits left inline.

_Assumption: the app wasn't built or run, so the centering improvement itself isn't verified — only the mechanism and its blast radius._

blotztask-mobile/src/feature/badge/components/badge-share-card.tsx
FFFLAREEE: This is intentional and follows the ticket scope. The badge detail DTO does not include an earned date, and the ticket explicitly asks us to make the share-card date optional and render detail-page shares without a date, with no backend changes. I’ll reword the PR description to make it clear that the date is intentionally omitted on the detail-page path rather than conditionally unavailable.

blotztask-mobile/src/feature/badge/components/badge-share-card.tsx
FFFLAREEE: Thanks for flagging this. I agree the detail endpoint does not currently verify that the user has earned the requested badge. The current wall only links to earned badges, and this ticket is explicitly frontend-only with no backend changes, so I’ll leave this unchanged here. It would be better tracked separately if locked badges become directly accessible.

blotztask-mobile/src/feature/badge/screens/badge-details-screen.tsx
FFFLAREEE: Confirmed. I’ll keep the release note scoped specifically to sharing: “Badge rewards can now be shared directly from their detail page.” The pre-existing Equip reward stub is explicitly out of scope for this ticket.

blotztask-mobile/src/feature/badge/screens/badge-details-screen.tsx
FFFLAREEE: Agreed. I’ll keep this PR focused and leave the hook rename and toast-key cleanup for a follow-up refactor.

LiChenyang-GZ:
image

LiChenyang-GZ:
image

lyx2-2 (approved):

PR #498, opened 2026-07-30
Add animated intro card to What's New carousel and update badge screenshot

lyx2-2:
## Summary

- Add animated intro card as the first slide of the What's New carousel: avatar2 with bob animation, sparkle stars (Reanimated), and a one-shot firework Lottie background
- Add `"intro"` card type to `WhatsNewCard` and a dedicated `IntroCard` component
- Update badge detail screenshot to the "篮以置信" achievement unlock screen
- Add `intro` copy in zh/en i18n files

## Release note

> none

**Status:**


.claude/skills/generate-whatsnew/blotz-whatsnew-intro.html
```diff
@@ -0,0 +1,511 @@
+<!DOCTYPE html>
```
sol-wizard: **Major:** this design prototype shouldn't be committed. The `generate-whatsnew` skill explicitly says the preview artifact is never committed, and `.gitignore:22` ignores `blotz-whatsnew.html` — this file only slipped through because of the `-intro` suffix.

It's a throwaway Option A/B mockup and the chosen copy already lives in the i18n files. Suggest deleting it here and widening the ignore to `.claude/skills/generate-whatsnew/blotz-whatsnew*.html` so the next variant can't leak either.

blotztask-mobile/src/feature/whats-new/screens/whats-new-screen.tsx
```diff
@@ -1,8 +1,113 @@
...
+    transform: [{ translateY: bobY.value }],
+  }));
+
+  return (
+    <View className="flex-1 items-center justify-center overflow-hidden px-6 pb-2">
+      <View className="absolute inset-0" pointerEvents="none">
+        <LottieView
+          source={LOTTIE_ANIMATIONS.fireworkTask}
+          autoPlay
+          loop={false}
```
sol-wizard, line 86: `autoPlay` + `loop={false}` fires once at mount, and all 7 cards stay mounted in the FlatList. The carousel has a back chevron and swipe-back, so returning to card 0 shows a static background with no firework.

Either loop it, or key the `LottieView` on the active card index so it remounts and replays.

blotztask-mobile/src/feature/whats-new/content/cards.ts
```diff
@@ -19,9 +19,22 @@ type AvatarCard = {
   bodyKey: string;
 };
 
-export type WhatsNewCard = ScreenshotCard | AvatarCard;
+type IntroCard = {
```
sol-wizard, line 22: nit — ignore if you like: `IntroCard` is structurally identical to `AvatarCard`; only the discriminant differs. Could be one type with a variant field. Keeping them separate is fine if intro is expected to diverge.

blotztask-mobile/src/feature/whats-new/screens/whats-new-screen.tsx
```diff
@@ -1,8 +1,113 @@
...
+    );
+    scale.value = withDelay(
+      delay,
+      withRepeat(
+        withSequence(withTiming(1.25, { duration: 900 }), withTiming(0.8, { duration: 900 })),
+        -1,
+        false
+      )
+    );
+  }, []);
```
sol-wizard: nit: empty dep array omits `delay` — will trip `react-hooks/exhaustive-deps` as a warning. Harmless today since `delay` is a constant per instance.

blotztask-mobile/src/i18n/locales/en/whats-new.json
```diff
@@ -4,6 +4,10 @@
     "continue": "Continue",
     "finish": "Let's go!"
   },
+  "intro": {
+    "title": "Guess what we've been up to.",
+    "body": "Swipe to find out～"
```
sol-wizard: nit: `～` is the fullwidth CJK tilde — reads a bit off in English copy. `Swipe to find out!` or no trailing mark would be more natural.

sol-wizard (commented):
🤖 **AI-generated PR review** — treat these as suggestions, use your own judgement, and verify before changing anything.

Overall this is clean and well scoped: the intro card follows the existing card-type pattern, both locales are added, and the screenshot swap is a real improvement — the old image leaked a raw enum (`类别：TaskCompletion`) into a user-facing promo shot.

One major item (inline, the committed prototype HTML) plus a few things that can't be anchored to the diff:

**Minor — `WHATS_NEW_VERSION` still `"2026-07"`** (`cards.ts:5`)
The seen-flag is `whats_new_seen_2026-07`. Anyone who already dismissed this version — e.g. an internal tester on a build that had these cards — will never see the new intro card. Fine if 2026-07 hasn't reached anyone yet; worth confirming.

**Minor — release-note status looks wrong**
"Hidden in production (feature-flagged)" is ticked, but What's New isn't flagged: `(protected)/index.tsx` renders it for every onboarded user whose seen-flag is unset. If the intent is "don't announce because this *is* the announcement surface", "Internal only" fits better.

**Minor — screenshot is 638px wide (~262 KB)**
The `generate-whatsnew` skill says scale to ~480px (`sips --resampleWidth 480`); the previous image was 480.

**nit — feel free to ignore:** the five sparkle loops and the bob animation run for the whole session even while the user is on a later card (all 7 stay mounted). Free to fix if you already key on the active index for the Lottie comment below.

No tests needed here — this is presentational and there's no logic worth pinning.

blotztask-mobile/src/feature/whats-new/content/cards.ts
lyx2-2: Ignore, still seperate, IntroCard already maps to a dedicated component with animations and effects that AvatarCard doesn't have.

blotztask-mobile/src/feature/whats-new/screens/whats-new-screen.tsx
lyx2-2: I think no need, firework animation only needs to be played once

blotztask-mobile/src/i18n/locales/en/whats-new.json
lyx2-2: changed to !

sol-wizard (approved):

PR #500, opened 2026-07-31
Block saving a task whose end time is after its deadline

AlexisEvan:
## Summary

- `hasDeadlineWarning` → `hasDeadlineConflict`, now wired into the task form's zod schema as a `.refine`, so a task whose end time falls after its deadline can no longer be saved (previously it was only a warning).
- `DeadlineSection` drops the one-shot toast for a persistent inline red error under the deadline row, so the message stays visible while the values are still invalid.
- Reworded the EN/ZH message to state the rule rather than the symptom.

## Release note

> You can no longer save a task whose end time is after its deadline — the deadline field now shows a clear error until you fix it.

**Status:**

### DIFF
--- a/blotztask-mobile/src/feature/task-add-edit/models/task-form-schema.ts
+++ b/blotztask-mobile/src/feature/task-add-edit/models/task-form-schema.ts
@@ -1,6 +1,6 @@
 import { z } from "zod";
 import { combineDateTime } from "../util/combine-date-time";
-import { isBefore, isEqual, startOfDay } from "date-fns";
+import { isAfter, isBefore, isEqual, startOfDay } from "date-fns";
 
 export const recurrenceValues = [
   "never",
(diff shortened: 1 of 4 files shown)


AlexisEvan:
截屏2026-07-31 下午7 07 20

截屏2026-07-31 下午7 08 25

When a user attempts to set a reminder after it's deadline , a red warning appears on the page and the "Create Task" button becomes unclickable unless the user modifies the time range to an allowed timeframe.

blotztask-mobile/src/feature/task-add-edit/models/task-form-schema.ts
```diff
@@ -94,22 +95,30 @@ export const taskFormSchema = z
       message: "recurrence.endDateBeforeStart",
       path: ["recurrenceEndDate"],
     },
-  );
+  )
+  .refine((data) => !hasDeadlineConflict(data), {
```
sol-wizard, line 99: Small thing: `path: ["deadlineTime"]` does nothing right now. Nothing in the app reads `formState.errors.deadlineTime`. `deadline-section.tsx` calls `hasDeadlineConflict` again and renders its own text.

It works, because both use the same function. But the same rule is now written in two places. If you read the error from the form instead, the message lives only in the schema. `task-form.tsx` already does this for `title` and `endTime`:

```tsx
{formState.errors.deadlineTime && (
  <Text className="text-red-500 text-sm mt-2 font-baloo">
    {t(formState.errors.deadlineTime.message || "")}
  </Text>
)}
```

Not important. Fine to leave as it is.

sol-wizard (commented):
🤖 AI-generated PR review — treat these as suggestions, use your own judgement and verify before changing anything.

sol-wizard (approved):

PR #503, opened 2026-08-01
Show start and end time on the calendar task card

LiChenyang-GZ:
## Summary

- Calendar task cards now show a time range under the title instead of just a date.
  - Same day: `19:00-21:00 · Today`
  - Across days: `19:00 · 05/15 - 20:00 · 05/16`
- Replaces `formatDateRange` with `formatTaskCardTimeRange`. The old util returned
  an empty string whenever the task fell on the selected day, so most cards showed
  no time information at all.
- The label no longer depends on `selectedDay` — it is derived purely from the
  task's own start/end time.
- Reuses the existing `calendar:header.today` key; no new translation strings.

## Release note

> none

**Status:**


blotztask-mobile/src/feature/calendar/util/format-task-card-time-range.ts
```diff
@@ -0,0 +1,33 @@
...
+}) => {
+  if (!startTime || !endTime) return "";
+
+  const start = parseISO(startTime);
+  const end = parseISO(endTime);
+
+  // Same day: "19:00-21:00 · Today"
+  if (isSameDay(start, end)) {
+    const timeRange =
+      startTime === endTime
```
sol-wizard: This compares the raw strings that came from the API. You have already parsed both dates two lines above, so compare the timestamps instead. It gives the same result today and still works if the server ever formats the two fields slightly differently:

```ts
start.getTime() === end.getTime()
```

`BaseTaskDTO` also has `timeType`, so `task.timeType === TaskTimeType.Single` is the direct way to ask the same question if you would rather pass it in from the card.

sol-wizard (commented):
🤖 AI-generated PR review — treat these as suggestions, use your own judgement and verify before changing anything.

sol-wizard (approved):

blotztask-mobile/src/feature/calendar/util/format-task-card-time-range.ts
LiChenyang-GZ: Good catch — fixed in 2786a161.

Went with `isSameMinute(start, end)` on the parsed dates rather than `getTime()`. Same fix for the serialisation issue you raised, but it also matches how the app itself defines a single-time task: `getTimeType` in `time-convertion.ts` uses `isSameMinute`, and the card renders `HH:mm`. Exact timestamp equality would render `19:00-19:00` for a task whose end is 30s off.

Kept it derived from the two values being displayed rather than taking `timeType` as a param, so the label can never disagree with what the card shows if a row ever has `timeType` out of sync with its times.

PR #504, opened 2026-08-01
feat(badges): add an equip endpoint for the achievement preview

AlexisEvan:
## Summary

Backend only — `POST /api/Badge/{id}/equip` lets a user pick which earned badge shows in the settings Achievement preview. Mobile is a separate PBI, so nothing calls it yet.

- Writes the existing `UserBadge.DisplayOrder` column as an ordered 3-slot queue: already equipped → no-op; fewer than 3 equipped → next free slot; exactly 3 → `[A,B,C]` + `D` → `[B,C,D]`.
- Selection only — no un-equip, so the preview always holds three badges.
- Unearned badge → 404. All shifts happen in one `SaveChangesAsync`.
- `BadgeDto.EquippedSlot` (`int?`) added from `ub.DisplayOrder`; `BadgeDto.DisplayOrder = index` is untouched, so badge wall sorting is unaffected.
- No migration — `DisplayOrder` already exists, and the award-time auto-assignment from #468 stays as the default.

## Release note

> none

**Status:**


AlexisEvan:
截屏2026-08-02 上午1 08 20
截屏2026-08-02 上午1 08 53
截屏2026-08-02 上午1 09 25

when add a fourth badge, remove the oldest badge, and make the second badge the first, which meet the FIFO

AlexisEvan:
截屏2026-08-02 上午1 13 42

blotztask-api/Modules/Badges/Controllers/BadgeController.cs
```diff
@@ -31,5 +33,17 @@ public async Task<BadgeByIdItemDto> GetBadgeById(int id, CancellationToken ct)
         var query = new GetBadgeByIdQuery { BadgeId = id, UserId = userId };
         return await getBadgeByIdQueryHandler.Handle(query, ct);
     }
+
+    // Equipping is idempotent: a retried request re-equips the same badge into the same slot
+    // instead of toggling it back off.
+    [HttpPost("{id}/equip")]
+    public async Task<List<EquippedBadgeDto>> EquipBadge(int id, CancellationToken ct)
```
Guotai812: i think we dont need return a list here, we need to have a get endpoint to return them anyway

blotztask-api/Modules/Badges/Commands/EquipBadge.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace BlotzTask.Modules.Badges.Commands;
+
+public class EquipBadgeCommand
+{
+    public required Guid UserId { get; init; }
+    public required int BadgeId { get; init; }
+}
+
+public class EquippedBadgeDto
+{
```
Guotai812: dont need this as well

blotztask-api/Modules/Badges/Commands/EquipBadge.cs
```diff
@@ -0,0 +1,70 @@
...
+        await db.SaveChangesAsync(ct);
+
+        logger.LogInformation(
+            "User {UserId} equipped badge {BadgeId} into slot {Slot}",
+            command.UserId, command.BadgeId, target.DisplayOrder);
+
+        return ToDto(equipped);
+    }
+
+    private static List<EquippedBadgeDto> ToDto(List<UserBadge> equipped) =>
```
Guotai812: also this one

AlexisEvan:
截屏2026-08-07 下午7 13 42

Equipped but does not return a list, displaying a 204.

截屏2026-08-07 下午7 37 00

blotztask-api/Modules/Badges/Commands/EquipBadge.cs
```diff
@@ -0,0 +1,56 @@
...
+    public async Task Handle(EquipBadgeCommand command, CancellationToken ct = default)
+    {
+        var userBadges = await db.UserBadges
+            .Where(ub => ub.UserId == command.UserId)
+            .ToListAsync(ct);
+
+        var target = userBadges.FirstOrDefault(ub => ub.BadgeId == command.BadgeId)
+            ?? throw new NotFoundException($"Badge with ID {command.BadgeId} not found for the current user.");
+
+        var equipped = userBadges
```
sol-wizard: nit, feel free to ignore: this runs before the already-equipped check below, so the no-op path builds a list it never uses. Moving the `target.DisplayOrder is not null` return above it reads a little cleaner.

blotztask-api/Modules/Badges/Queries/GetUserBadges.cs
```diff
@@ -40,7 +41,8 @@ public async Task<List<BadgeDto>> Handle(GetUserBadgesQuery query, CancellationT
                 Id = badge.Id,
                 Name = language == Language.Zh ? badge.NameZh : badge.NameEn,
                 IconUrl = badge.IconUrl,
-                DisplayOrder = index
+                DisplayOrder = index,
```
sol-wizard, line 44: These two are easy to mix up: `DisplayOrder` here is just the position in the list, while `EquippedSlot` is the real DB column. Maybe a one-line comment on each so nobody picks the wrong one later.

blotztask-api/Modules/Badges/Queries/GetUserBadges.cs
```diff
@@ -29,7 +29,8 @@ public async Task<List<BadgeDto>> Handle(GetUserBadgesQuery query, CancellationT
                     badge.IconUrl,
                     badge.NameEn,
                     badge.NameZh,
-                    ub.EarnedAtUtc
+                    ub.EarnedAtUtc,
+                    EquippedSlot = ub.DisplayOrder
```
sol-wizard, line 33: `DisplayOrder` only ever gets set for a user first 3 badges (`AwardNewBadgesToUser`, `order <= 2`), and existing rows were never backfilled, so a lot of users have `null` in all of them. Worth deciding whether this PR should add a backfill.

blotztask-api/Modules/Badges/Queries/GetUserBadges.cs
```diff
@@ -55,4 +57,5 @@ public class BadgeDto
     public required string Name { get; set; }
     public required string IconUrl { get; set; }
     public int DisplayOrder { get; set; }
+    public int? EquippedSlot { get; set; }
```
sol-wizard, line 63: Worth returning this from `GetBadgeById` too, so a single badge lookup can also tell whether it is already equipped.

blotztask-api/Modules/Badges/Commands/EquipBadge.cs
```diff
@@ -0,0 +1,56 @@
...
+    public required int BadgeId { get; init; }
+}
+
+public class EquipBadgeCommandHandler(BlotzTaskDbContext db, ILogger<EquipBadgeCommandHandler> logger)
+{
+    private const int SlotCount = 3;
+
+    public async Task Handle(EquipBadgeCommand command, CancellationToken ct = default)
+    {
+        var userBadges = await db.UserBadges
```
sol-wizard: Read-modify-write with nothing guarding it. Two equips in flight (double tap, or two devices) both read this list and both renumber, so you can end up with duplicate slots or four equipped.

The `while` below self-heals on the next equip, so the impact is a briefly wrong preview rather than corruption. Fine to leave as is; a filtered unique index on `(UserId, DisplayOrder)` would close it properly if you want it airtight.

blotztask-api/Modules/Badges/Commands/EquipBadge.cs
```diff
@@ -0,0 +1,56 @@
...
+            .ToList();
+
+        if (target.DisplayOrder is not null)
+        {
+            logger.LogInformation(
+                "Badge {BadgeId} is already on display for user {UserId}", command.BadgeId, command.UserId);
+            return;
+        }
+
+        // FIFO: the display holds at most three badges, so a new one pushes the oldest out.
```
sol-wizard, line 38: Was "no un-equip" in the PBI, or your own call? With FIFO the user cannot pick which badge gets replaced, so just checking that is intended.

blotztask-api/Modules/Badges/Commands/EquipBadge.cs
```diff
@@ -0,0 +1,56 @@
...
+        while (equipped.Count >= SlotCount)
+        {
+            equipped[0].DisplayOrder = null;
+            equipped.RemoveAt(0);
+        }
+
+        equipped.Add(target);
+
+        for (var i = 0; i < equipped.Count; i++)
+            equipped[i].DisplayOrder = i;
```
sol-wizard, line 48: Up to you, but a quick test with `TestDbContextFactory` would cover this: equip 3, equip a 4th, check the oldest went back to `null` and the rest are 0, 1, 2. There are no badge tests yet.

blotztask-api/Modules/Badges/Controllers/BadgeController.cs
```diff
@@ -31,5 +33,17 @@ public async Task<BadgeByIdItemDto> GetBadgeById(int id, CancellationToken ct)
         var query = new GetBadgeByIdQuery { BadgeId = id, UserId = userId };
         return await getBadgeByIdQueryHandler.Handle(query, ct);
     }
+
+    [HttpPost("{id}/equip")]
```
sol-wizard, line 37: Checklist: this is a new feature that just is not wired up yet, so "Hidden in production" fits better than "Internal only".

blotztask-api/Modules/Badges/Commands/EquipBadge.cs
sol-wizard: yes it is intended

sol-wizard (changes requested):

sol-wizard (approved):

PR #505, opened 2026-08-02
Show an empty state when the AI finds no tasks, instead of an error

Guotai812:
## Summary

The backend expressed "the user said nothing task-shaped" by throwing (`AiErrorCode.NoTasksExtracted`), so a valid empty outcome arrived on the error channel: red toast, transcript wiped, blank screen. Users could not tell a mis-hearing from a genuinely task-free sentence, or keep what they said.

- **Backend:** an empty extraction now returns a normal `AiGenerateMessage` with `isEmptyResult`. Genuine faults (cancel, 429, content filter, unhandled) still throw.
- **Why a flag, not empty lists:** the `extracted*` lists are cumulative for the whole session, so a barren turn can still return a non-empty snapshot. `ToolCallCount` is the only per-turn signal.
- **Mobile:** the sheet shows what was heard plus a "Save as note" action; drafts from earlier turns stay put.
- Empty results stop counting toward `ai_task_generation_failed`, which they were 59% of.

Backend and app must ship together: older app versions get no empty-state feedback.

## Release note

> When the AI doesn't find any to-dos in what you said, you now see what it heard and can save it as a note, instead of a red error message.

**Status:**


blotztask-mobile/src/feature/ai-task-generate/screens/ai-task-sheet-screen.tsx
```diff
@@ -80,7 +83,7 @@ export default function AiTaskSheetScreen() {
   );
   const { addTaskAsync, isAdding, createRecurringTaskAsync, isCreatingRecurringTask } =
     useTaskMutations();
-  const { createNoteAsync, isNoteCreating } = useNotesMutation();
+  const { createNote, createNoteAsync, isNoteCreating } = useNotesMutation();
```
nxn-nicole: why use createnote instead of createNoteAsync

blotztask-mobile/src/feature/ai-task-generate/hooks/useAiTaskGenerator.ts
```diff
@@ -178,6 +181,9 @@ export function useAiTaskGenerator({
     pendingInputModeRef.current = null;
     const inputText = result.userInput;
 
+    // A turn that extracted nothing is a normal result, so it arrives here rather than as an error.
+    setEmptyResult(result.isEmptyResult ? { userInput: inputText ?? "" } : null);
```
nxn-nicole: looks like the empty result is just use input? then why do you need another state instead of just using the transcript?

blotztask-mobile/src/feature/ai-task-generate/models/ai-result-message-dto.ts
```diff
@@ -3,6 +3,9 @@ export interface AiResultMessageDTO {
   extractedTasks?: ExtractedTaskDTO[];
   extractedNotes?: AiNoteDTO[];
   extractedRecurringTasks?: ExtractedRecurringTaskDTO[];
+  // This turn called no tools, so nothing was extracted from what the user said. The extracted*
+  // lists are cumulative for the session, so their length cannot tell us this on its own.
```
nxn-nicole: check all the comments again. they look like ai generated and doesn't contain much useful info

nxn-nicole (changes requested):

nxn-nicole (approved):

PR #512, opened 2026-08-08
update build shell script

sol-wizard:
## Summary

Three unrelated pieces of internal work.

**Task ownership (security).** Task completion, subtask completion and subtask delete looked rows up by ID alone — any authenticated user could toggle or delete another user's data by enumerating integer IDs, and completion awarded the badge to the caller. Each command now carries `UserId` and filters on it (subtasks via `ParentTask.UserId`), matching `EditTask`. Still unchecked, for a follow-up: `GetSubtasksById`, `UpdateSubtask`, `AddSubtask`, `ReplaceSubtasks`.

**build.ps1.** Fails early if the EAS CLI is missing, only rewrites `eas.json` when a value changed, restores it on build failure, and sets `EXPO_NO_CAPABILITY_SYNC` so EAS stops overwriting manually-managed iOS capabilities.

**app.config.js.** `usesAppleSignIn: true`; `appleTeamId` now reads `EXPO_APPLE_TEAM_ID`.

Plus `create-blotz-pbi` skill updates. Tests fail 85/98 here — identical on `main`, pre-existing.

## Release note

> _none_

**Status:**

### DIFF
--- a/.claude/skills/create-blotz-pbi/SKILL.md
+++ b/.claude/skills/create-blotz-pbi/SKILL.md
@@ -15,7 +15,71 @@ A PBI is ready only when it has all three:
 
 **Exception — POC / investigation PBI:** the goal is to *find* the solution, so #2 is replaced by a clear **investigation goal + what success looks like** (questions to answer, decision to reach). #1 and #3 still apply. Ask whether it's a POC if unclear.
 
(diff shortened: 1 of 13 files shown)


lyx2-2 (approved):
👍

PR #515, opened 2026-08-12
Privacy statement

LiliZ3681:
## Summary

Sync the in-app Privacy Policy by opening the privacy policy from the website in the in-app browser.

## Release note

> _Updated the in-app Privacy Policy to the latest version._

**Status:**

https://github.com/user-attachments/assets/68bf419e-be94-41c1-b7bd-15ceda99e424


blotztask-mobile/src/feature/settings/screens/settings-about-screen.tsx
```diff
@@ -3,24 +3,33 @@ import { SafeAreaView } from "react-native-safe-area-context";
...
     Linking.openURL(WEBSITE_URL);
   };
 
   const handlePrivacyPolicy = () => {
-    router.push("/settings/privacy-policy");
+    const privacyPolicyUrl = i18n.resolvedLanguage?.startsWith("zh")
+      ? PRIVACY_POLICY_URLS.zh
+      : PRIVACY_POLICY_URLS.en;
+
+    void WebBrowser.openBrowserAsync(privacyPolicyUrl);
```
sol-wizard: `void` doesn't suppress an unhandled rejection, it just discards the promise. `openBrowserAsync` rejects on Android if a browser session is already presented (fast double-tap on the row), and also if the device has no browser — so this can surface as an unhandled rejection with no visible result for the user.

```ts
WebBrowser.openBrowserAsync(privacyPolicyUrl).catch(() => Linking.openURL(privacyPolicyUrl));
```

That swallows the error and still gets the user to the policy.

sol-wizard (commented):
**AI-generated PR review** — treat these as suggestions, use your own judgement and verify before changing anything.

Checked the swap from the bundled privacy policy screen to the hosted one:

- Both `blotztask.com/en/privacy/` and `/zh/privacy/` return 200 (the un-prefixed `/privacy/` is a 404, and the code always supplies a locale, so that's fine).
- No dangling references left — route, screen, the `privacyPolicy` i18n namespace and both JSON files are removed together, and nothing else imported them.
- `expo-web-browser` was already a dependency, so no native rebuild needed.
- The language pick matches the app's own device-language mapping in `src/i18n/index.ts`, and an undefined `resolvedLanguage` falls back to `en`.

One minor comment below. No blocking issues.

sol-wizard (approved):

PR #521, opened 2026-08-22
ci: one-button staging release for API + iOS + Android

LiChenyang-GZ:
## Summary

Turns "release to staging" into one `workflow_dispatch` button, replacing the local `build.ps1` run.

- **New `deploy-staging.yml`** — version bump picked from a dropdown (no free text), then: compute version → build iOS + Android in parallel with `eas build --auto-submit` → deploy the API → check `/health` → commit the version back to `main`. A failed app build stops the API deploy.
- **`deploy-staging-api.yml`** — `on: push` removed, `workflow_call` added, so the staging API no longer deploys on merges to `main`; it now ships with the app. `workflow_dispatch` is kept for API-only redeploys.
- **`eas.json`** — Android submit track `alpha` → `internal`; preview `APP_VERSION` aligned to `1.3.0` (was iOS `1.2.1`, Android `1.0.3`).

The platforms are separate matrix legs, so one failing can be retried with "Re-run failed jobs" without rebuilding the other.

Needs an `EXPO_TOKEN` repository secret before the first run.

## Release note

> none

**Status:**


dksjl661 (approved):

.github/workflows/release-staging.yml
sol-wizard, line 1: i will name this file better to represent the full staging deployment is here

.github/workflows/deploy-staging.yml
```diff
@@ -0,0 +1,185 @@
+name: Staging - Release app and API
+
+# One button for a full staging release. Runs in this order, and the order matters:
```
sol-wizard: this is ai generated , if you think important keep them in below each action step

.github/workflows/deploy-staging.yml
```diff
@@ -0,0 +1,185 @@
...
+            exit 0
+          fi
+
+          git config user.name  "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add blotztask-mobile/eas.json
+          git commit -m "chore: staging release $NEW"
+          # main may have moved during the 10-20 minute build
+          git pull --rebase origin main
+          git push origin main
```
sol-wizard: **Problem:** `main` has branch protection — 1 approving review required, no bypass list. `github-actions[bot]` is not exempt, so this `git push origin main` will be rejected. You can check: no commit on `main` has ever come from the bot.

**Why it matters:** by this point both apps are already submitted and the API is deployed. The run then fails on the very last step, the version bump never lands, and the next run computes the same version again and ships it a second time — the drift this workflow is meant to remove.

**Fix:** replace this hand-rolled block (commit, pull --rebase, push) with a maintained action. Two options:
- `peter-evans/create-pull-request` — opens a PR with the bump; one click to merge; works with branch protection as-is. Simplest.
- `stefanzweifel/git-auto-commit-action` + a PAT from an admin stored as a secret and passed to `actions/checkout` as `token:` — fully automatic, works because `enforce_admins` is off.

Either way the same `jq` version-write is also duplicated from the build job above; worth keeping in one place.

---
_Note: the suggested fix above is AI-generated and I have not verified it myself. Please check it actually works before relying on it._

.github/workflows/release-staging.yml
```diff
@@ -0,0 +1,185 @@
...
+      - name: Build and submit
+        env:
+          EXPO_TOKEN: ${{ secrets.EXPO_TOKEN }}
+          # Apple Developer capabilities are managed by hand, same as build.ps1
+          EXPO_NO_CAPABILITY_SYNC: '1'
+        run: |
+          eas build \
+            --platform ${{ matrix.platform }} \
+            --profile preview \
+            --auto-submit \
```
sol-wizard, line 130: **Problem:** `--auto-submit` needs store credentials. `build.ps1` worked because it ran on a Mac where `keys/` exists. The runner has no `keys/` (it is gitignored), yet the preview profile sets `EXPO_ASC_API_KEY_PATH: ./keys/AuthKey_….p8`, and the Android submit profile has no `serviceAccountKeyPath` either.

**Why it matters:** if EAS does not already hold these keys, the submit fails — but only after the 15-minute build has finished, so you pay for the build and still get nothing in TestFlight / Play.

**Check before the first run:** `eas credentials -p ios` and `-p android` should show an App Store Connect API key and a Play service-account key stored remotely. If they are, the `EXPO_ASC_API_KEY_PATH` env can be dropped from eas.json to avoid confusion.

---
_Note: the suggested fix above is AI-generated and I have not verified it myself. Please check it actually works before relying on it._

.github/workflows/release-staging.yml
```diff
@@ -0,0 +1,185 @@
...
+permissions:
+  contents: write
+  id-token: write
+
+concurrency:
+  group: deploy-staging
+  cancel-in-progress: false
+
+jobs:
+  version:
```
sol-wizard, line 31: **Expectation:** a staging release should only ever be built from `main`. Nothing from a feature branch should reach staging.

**Problem:** the "Run workflow" dropdown lets you pick any branch that has this file. The build then uses that branch's code, but the last job always writes the version to `main`. So staging can end up running a feature branch while `main` says it shipped that version — nobody can tell from git what is actually on staging.

**Fix:** one line on this job: `if: github.ref == 'refs/heads/main'`. A dispatch from any other branch then does nothing instead of half-releasing. (An alternative is a "Deployment branches: main only" rule on the `Staging` environment, but that only guards the API deploy job, not the EAS build, so the `if:` is the simpler complete option.)

---
_Note: the suggested fix above is AI-generated and I have not verified it myself. Please check it actually works before relying on it._

blotztask-mobile/eas.json
```diff
@@ -105,7 +105,7 @@
         "ascAppId": "6761295231"
       },
       "android": {
-        "track": "alpha"
+        "track": "internal"
```
sol-wizard, line 108: **Problem:** `alpha` (closed testing) and `internal` (internal testing) are two separate tester lists in Play Console. Switching the track does not move testers across.

**Why it matters:** everyone on the alpha list stops receiving staging builds after the first run, and the internal list is capped at 100 testers.

**Check before the first run:** open Play Console → Testing → Internal testing and make sure the current staging testers are on that list.

---
_Note: the suggested fix above is AI-generated and I have not verified it myself. Please check it actually works before relying on it._

.github/workflows/release-staging.yml
```diff
@@ -0,0 +1,185 @@
...
+
+  verify-and-record:
+    name: Health check and record version
+    needs: [version, build, deploy-api]
+    runs-on: ubuntu-latest
+    steps:
+      # Retries because the Azure Web App takes a moment to come back up after a
+      # deploy. /health also covers the database, which matters here because the
+      # deploy runs EF migrations immediately beforehand.
+      - name: Check the staging API is healthy
```
sol-wizard, line 177: **Idea:** after the API deploy, we can hit `/health` to confirm the deploy + EF migration actually produced a working API. `/health` also checks the database, which matters because the migration runs right before it.

**Do we need it?** I think yes. Without it the run goes green even if the API is down after the migration, and testers find out on the app instead. PBI #1561 also lists "a failing health check fails the run" as an acceptance criterion.

**How:** what is here is the right shape — `curl` with retries to ride out the App Service warm-up. Two things to check before calling it good:
- (a) 6 × 10s = 60s may be tight for a cold .NET start on App Service. `--retry 12 --retry-delay 15` (~3 min) gives more room.
- (b) Order: by this point both apps are already submitted with the new version. If this check fails, the version is never written to eas.json even though it has shipped. Commit the version first, then health-check — a red check should mean "API is down", not "eas.json is now wrong".

**Is the current approach good?** Yes, with those two tweaks.

---
_Note: the suggested fix above is AI-generated and I have not verified it myself. Please check it actually works before relying on it._

.github/workflows/deploy-staging.yml
```diff
@@ -0,0 +1,185 @@
...
+    name: Work out the new version
+    runs-on: ubuntu-latest
+    outputs:
+      new_version: ${{ steps.bump.outputs.new_version }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Bump the preview version
+        id: bump
+        working-directory: blotztask-mobile
```
sol-wizard: nit: almost every `run` step in this file is in `blotztask-mobile`. You can set it once at the top of the workflow:

```yaml
defaults:
  run:
    working-directory: blotztask-mobile
```

and drop the per-step / per-job copies. One catch: the health-check step in the last job runs **before** `actions/checkout`, so the directory would not exist yet and that step would fail. Move the checkout above it (which you want anyway, see the health-check comment) and it works.

---
_Note: the suggested fix above is AI-generated and I have not verified it myself. Please check it actually works before relying on it._

sol-wizard (approved):

sol-wizard (changes requested):
Before merging, please dry-run the two things that will break on the first real run, since this workflow cannot be tested end-to-end without doing an actual staging release:

1. **Store credentials:** run `eas build --platform ios --profile preview --auto-submit --non-interactive` (and android) from a folder **without** `keys/`, using `EXPO_TOKEN`. This proves the submit works with the keys stored in EAS, which is what the runner will have.
2. **Push to main:** confirm how the version commit gets onto `main` given branch protection (see the inline comment) — either the PR-based action or a PAT — and test that path once.

The rest of the inline comments are smaller. Note the suggested fixes in them are AI-generated and not verified — please check them yourself.

.github/workflows/deploy-staging.yml
LiChenyang-GZ: OK

.github/workflows/deploy-staging.yml
LiChenyang-GZ: OK

sol-wizard (approved):

.github/workflows/release-staging.yml
LiChenyang-GZ: OK

.github/workflows/release-staging.yml
LiChenyang-GZ: OK

.github/workflows/deploy-staging.yml
LiChenyang-GZ: OK

.github/workflows/release-staging.yml
LiChenyang-GZ: we dont need EXPO_ASC_API_KEY_PATH

blotztask-mobile/eas.json
LiChenyang-GZ: no people test android now

.github/workflows/release-staging.yml
LiChenyang-GZ: ok

PR #523, opened 2026-08-23
fix: stabilize screen-entry animations

FFFLAREEE:
## Summary

- Prevent initial async data and mount-time animations from shifting the New Task and Settings layouts.
- Keep existing notes static when opening Notes while preserving animations for newly created notes.
- Smooth user-driven Repeat layout changes and make shared divider animations opt-in.

## Release note

> Task forms, notes, and settings now open without distracting layout jumps or slide-in animations.

**Status:**


FFFLAREEE:
https://github.com/user-attachments/assets/33af0604-5f76-414f-9720-96f5f00e6c3c

blotztask-mobile/src/feature/task-add-edit/task-form.tsx
```diff
@@ -197,35 +208,40 @@ const TaskForm = ({ mode, dto, recurrenceEditMode, onSubmit }: TaskFormProps) =>
...
+              formState.dirtyFields.recurrence ? MotionAnimations.upEntering : undefined
+            }
+            exiting={MotionAnimations.outExiting}
+            layout={MotionAnimations.layout}
+          >
             <FormDivider />
-          </>
+            <AlertSelect control={control} />
+          </Animated.View>
         )}
```
lyx2-2, line 242: When editing an already-saved, non-recurring task, shouldShowRecurrence evaluates to false (since recurrenceEditMode is "hidden" for tasks without a recurringOccurrence), so the entire "Repeat" section (RecurrenceSelect + RecurrenceEndSection) renders nothing. However, the FormDivider that used to sit after AlertSelect was moved to sit before it instead, while the unconditional FormDivider after DeadlineSection was left in place. With the Repeat section absent, these two dividers now render directly adjacent to each other with no content between them, producing a large blank gap with two faint dashed lines instead of the expected single divider.

lyx2-2: Image

blotztask-mobile/src/shared/components/form-divider.tsx
```diff
@@ -2,7 +2,15 @@ import { MotionAnimations } from "@/shared/constants/animations/motion";
...
 
-export const FormDivider = ({ marginVertical = 16 }: { marginVertical?: number }) => {
+type FormDividerProps = {
+  marginVertical?: number;
+  animateLayout?: boolean;
+};
+
+export const FormDivider = ({
+  marginVertical = 16,
+  animateLayout = false,
```
lyx2-2: It looks like this was meant to turn off the divider animation for one specific case (avoiding the animation firing on initial mount/page entry), but since animateLayout was changed as the default on a shared component instead of scoped to that one case, it ended up turning off the divider animation everywhere else too, including places outside this PR's scope that never had a mount-time animation problem to begin with.

You may consider defaulting animateLayout to true and opting out only the specific divider that had the mount-time jump issue, so this doesn't regress components outside this PR's scope.

blotztask-mobile/src/feature/task-add-edit/components/event-tab.tsx
```diff
@@ -20,7 +20,7 @@ import { MotionAnimations } from "@/shared/constants/animations/motion";
 import { formatLocalizedDate } from "@/shared/util/localized-date-format";
 import { combineDateTime } from "../util/combine-date-time";
 
-export const EventTab = () => {
+export const EventTab = ({ animateOnEnter = false }: { animateOnEnter?: boolean }) => {
```
lyx2-2: Defaulting animateOnEnter to false here also silently affects note-time-picker-sheet.tsx:142,146, which calls <ReminderTab /> / <EventTab /> without this prop.

You can go to the Notes tab, swipe a note left, tap "Add to task", this opens the time picker sheet with a Reminder/Event toggle. Switching between "Reminder" and "Event" there no longer slides the content in; it just pops in instantly. 

Image 
You can try to toggle between 提醒 and 事件 here, the animation here disappear because this "false"

blotztask-mobile/src/feature/task-add-edit/task-form.tsx
```diff
@@ -197,35 +208,40 @@ const TaskForm = ({ mode, dto, recurrenceEditMode, onSubmit }: TaskFormProps) =>
...
-          </>
+          </Animated.View>
         )}
 
         {!isRecurringTaskForm && (
-          <>
-            <AlertSelect control={control} />
+          <Animated.View
+            entering={
+              formState.dirtyFields.recurrence ? MotionAnimations.upEntering : undefined
```
lyx2-2: The Alert (reminder) section only shows an animation when it reappears if formState.dirtyFields.recurrence is true. But dirtyFields resets to false whenever the field's value goes back to its default — even if the user did change it before.

Steps to reproduce:
1. Open New Task. Repeat defaults to "Never" (Alert section is visible).
2. Change Repeat to "Daily", Alert section disappears.
3. Change Repeat back to "Never" — Alert section should reappear with a smooth animation.

What happens instead: No animation plays, it just pops back in instantly, because the value is back to the default, so dirtyFields.recurrence is false again, even though the user did change it.

Suggested fix: use a sticky flag instead (like hasChangedTab used elsewhere in this same file for the Reminder/Event tabs), so it stays true once the user has changed Repeat at all, instead of relying on dirtyFields.

lyx2-2 (changes requested):

lyx2-2 (approved):
👍

PR #528, opened 2026-08-28
add sentry span tracing

nxn-nicole:
## Release note

> _your one-line user-facing note here, or "none"_

**Status:**


blotztask-mobile/src/app/_layout.tsx
```diff
@@ -33,7 +33,7 @@ Sentry.init({
   enabled: process.env.EXPO_PUBLIC_APP_ENV !== "development",
   sendDefaultPii: true,
   enableNative: true,
-  replaysSessionSampleRate: 0.1,
+  tracesSampleRate: 0.7,
```
sol-wizard: `replaysSessionSampleRate` was **replaced**, not renamed — these are unrelated settings. Session Replay for normal sessions is now off (defaults to 0); only on-error replays survive. Looks accidental — keep both lines.

Also `0.7` is high: with no navigation integration, every API call becomes its own transaction, and the widget sync alone fires 8 requests per app open. Suggest `0.1`–`0.2`.

blotztask-mobile/src/shared/services/api/client.ts
```diff
@@ -1,7 +1,21 @@
...
 import { API_CONFIG } from "./config";
 import { setupRequestInterceptor, setupResponseInterceptor } from "./interceptors";
 
+// trace axios request with Sentry span to record request time
+function traced<T>(method: string, url: string, run: () => Promise<T>): Promise<T> {
+  return Sentry.startSpan(
+    {
+      // create fix path name
+      name: `${method} ${url.split("?")[0].replace(/\/\d+/g, "/:param")}`,
+      op: "http.client",
```
sol-wizard: Setting `tracesSampleRate` auto-enables `reactNativeTracingIntegration`, which already instruments XHR — and axios uses XHR in RN. That auto-span only fires when a parent span exists, so this wrapper is what makes it work; but each request now emits **two** spans, both with `op: "http.client"`.

Either give the outer span a different `op`, or add `Sentry.reactNavigationIntegration()` so requests nest under screen transactions the standard way.

sol-wizard (commented):
🤖 **AI-generated PR review** — treat these as suggestions, use your own judgement and verify before changing anything.

Two things worth a look (inline). The rest checks out:

- `getRawInstance()` removal is safe — no callers anywhere.
- `fetchTasksForDate(date, true)` → `(date)` fixes the **only** TypeScript error in `src/` on main (`TS2554: Expected 1 arguments, but got 2`). Good catch.
- Path normalization `/\d+/ → /:param` is correct for this codebase — all path IDs are numeric, no GUIDs, and query strings only carry dates/timezone IDs.

nit: description is still the raw template. A one-line summary would help, especially since 2 of the 3 files aren't about tracing.

sol-wizard (approved):

PR #533, opened 2026-09-03
feat(review): return tasksCompleted on the review report

AlexisEvan:
## Summary

`GET /api/review` now returns `tasksCompleted` — tasks whose `CompletedAt`
falls inside the resolved `ReviewPeriod`.

- The count is a `CountAsync` over `TaskItems`, written inline in both
  `GetReviewQueryHandler` and `GenerateReviewCommandHandler`. The task set
  loaded for the AI is wider (planned OR completed in the period), so it
  cannot be reused here.
- `GenerateReviewCommandHandler.Handle` now has one return site, so the count
  also covers the already-exists and concurrency-loser paths. That restructure
  is mostly re-indentation — `git diff -w` reduces it to ~29 lines.
- Weekly and monthly run the same query over their own `[StartUtc, EndUtc)`,
  with no branch on `PeriodType`.
- Computed per request: no new column, no EF migration.
- Tests cover monthly and weekly counts, the half-open period boundary, and a
  Sydney DST month whose window spans UTC+11 → UTC+10.

## Release note

> none

**Status:**


blotztask-api/Modules/Reviews/ReviewMetrics.cs
```diff
@@ -0,0 +1,24 @@
+using BlotzTask.Infrastructure.Data;
+using BlotzTask.Modules.Reviews.Domain;
+using Microsoft.EntityFrameworkCore;
+
+namespace BlotzTask.Modules.Reviews;
+
+public static class ReviewMetrics
```
sol-wizard: why we using static class here ? why not we query the data from the handler ?

blotztask-api/Modules/Reviews/Queries/GetReview.cs
```diff
@@ -48,12 +48,14 @@ public async Task<ReviewReportDto> Handle(
...
                 ct);
 
+        var tasksCompleted = await ReviewMetrics.CountCompletedAsync(db, query.UserId, period, ct);
+
         return new ReviewReportDto
         {
             PeriodType = period.PeriodType,
             PeriodStartLocal = period.StartLocalDate,
             PeriodEndLocalExclusive = period.EndLocalDateExclusive,
-            // letter / generatedAt stay null when no review exists yet.
```
sol-wizard, line 56: why remove the comment, the comment also looks like not too obvious please rewrite the comment

blotztask-api/Modules/Reviews/Commands/GenerateReview.cs
```diff
@@ -146,22 +146,29 @@ await recordAiUsageService.RecordAiUsageAsync(new RecordAiUsageRequest
...
-            return MapToDto(winningReport, period, threshold);
+            return await MapToDtoAsync(winningReport, period, threshold, command.UserId, ct);
         }
 
         logger.LogInformation(
             "Saved {PeriodType} review {ReportId} for user {UserId} ({StartLocal}, {TimeZoneId})",
             period.PeriodType, report.Id, command.UserId, period.StartLocalDate, period.TimeZoneId);
 
-        return MapToDto(report, period, threshold);
+        return await MapToDtoAsync(report, period, threshold, command.UserId, ct);
```
sol-wizard: your method is call map to dto but under the hood you in the fucntion you call a query

sol-wizard (dismissed):

blotztask-api/Modules/Reviews/ReviewMetrics.cs
sol-wizard: Direction after discussion: delete `ReviewMetrics.cs` and inline this count query in both handlers (`GetReview` and `GenerateReview`). We prefer each handler being self-contained here over a shared helper, even at the cost of a small duplication.

One consequence to cover: `ReviewMetricsTests` loses its target once the class is gone. Please retarget the same test cases at `GetReviewQueryHandler` — same seeded data, but assert on `dto.TasksCompleted`. The DST-window, tick-boundary, and user-isolation cases are worth keeping, and testing through the handler is a better test anyway since it covers what the API actually returns.

blotztask-api/Modules/Reviews/Commands/GenerateReview.cs
sol-wizard: Suggested fix: restructure `Handle` to a single return site — then the mapper disappears entirely instead of needing a rename:

- existing-report branch: `report = existing;` instead of returning
- race-loser catch: `report = winner;` instead of returning
- happy path: `report = newReport;`

Then, after the branches converge, run the completed-count query once (inline in the handler, per the `ReviewMetrics` thread) and build the `ReviewReportDto` inline at the end of `Handle`:

```csharp
var tasksCompleted = await db.TaskItems
    .AsNoTracking()
    .CountAsync(t => t.UserId == command.UserId
        && t.CompletedAt != null
        && t.CompletedAt >= period.StartUtc
        && t.CompletedAt < period.EndUtc, ct);

return new ReviewReportDto { ..., TasksCompleted = tasksCompleted, ... };
```

Result: the count runs exactly once on every path, the DTO is constructed in one place, and no method hides a query behind a mapping name. Please also move the "AI's task set is wider (planned OR completed)" note next to the new count query so it isn't lost.

blotztask-api/Modules/Reviews/Queries/GetReview.cs
sol-wizard: Please restore it — it still holds after this change (`report?.AiGeneratedLetter` is still null when no report exists) — and make it say why the DTO exists at all in that case, e.g.:

```csharp
// A report may not exist yet (not generated, or the period hasn't ended).
// The DTO is still returned with the metrics; letter/generatedAt stay null.
```

Guotai812:
there is no ReviewMetrics.CountCompletedAsync in ur code, you can update your summary

Guotai812 (approved):

AlexisEvan:
> there is no ReviewMetrics.CountCompletedAsync in ur code, you can update your summary

Done

PR #544, opened 2026-09-12
feat(review): split the AI letter into theme, body and next-period suggestion

AlexisEvan:
## Summary

Backend only. The weekly/monthly review letter is now generated as three parts — theme, body and one suggestion for next period — stored and returned separately so the app can lay them out as distinct blocks.

- The prompt returns JSON pinned to a strict schema; theme and suggestion are null for a quiet period
- `ReviewReport` gains nullable `Theme` / `OneThingToTryNext` (migration `SplitReviewLetterIntoParts`); `AiGeneratedLetter` keeps the body, so existing reports still load
- `ReviewReportDto` adds `theme` and `oneThingToTryNext`; `letter` keeps its name for clients already in the stores
- The last 3 themes of the same period type are fed back so consecutive letters don't repeat a theme
- A filtered or malformed response falls back to the whole text as the body instead of failing the request
- The app still renders `letter` only — showing three blocks needs a frontend follow-up

## Release note

> none

**Status:**


AlexisEvan:
截屏2026-09-12 下午9 07 55

blotztask-api/Modules/Reviews/Domain/ReviewLetter.cs
```diff
@@ -0,0 +1,50 @@
...
+                    parsed.Body.Trim(),
+                    Clean(parsed.Theme),
+                    Clean(parsed.OneThingToTryNext));
+            }
+        }
+        catch (JsonException)
+        {
+            // Fall through to the whole-response fallback below.
+        }
+
```
sol-wizard, line 36: (AI-assisted review, please use your own judgment.)

With strict schema the model won't return prose, so what reaches this line in practice is truncated JSON from `FinishReason == Length` or `ContentFilter` (both HTTP 200, so the 400 content_filter catch doesn't fire). That gets stored as the body, and since the handler returns the existing report forever, the user is stuck with `{"theme": "Exams", "body": "You spent the wee` as their letter.

Suggest: in `GenerateLetterAsync` check `response.Value.FinishReason == ChatFinishReason.Stop` before parsing, and throw `AiErrorCode.EmptyResponse` on a parse failure instead of saving raw text. The user then gets the existing "Couldn't generate, try again" toast and a clean retry. `Parse_JsonWithoutABody_FallsBackToTheWholeResponseAsBody` would flip to expect the throw.

LiliZ3681 (approved):

PR #545, opened 2026-09-14
Fix task checkbox not marking done instantly, recurring tasks getting stuck

LiliZ3681:
## Summary

Task checkbox didn't visually update (or update at all) on tap.

- Regular tasks: optimistic update wrote to the wrong cache key, so the checkbox only updated via a slow background refetch.
- Recurring tasks: same wrong-key bug in the invalidation call, but no fallback — could stay stuck indefinitely. Added the missing optimistic update here too.
- Recurring tasks also popped visually on first completion — completing one assigns it a real id, and the list's key extractor picked that over the item's stable recurring identity. Reordered the check.

## Release note

> Checking off a task, including recurring ones, now marks it done instantly with no lag or glitch.

**Status:**


sol-wizard:
🤖 AI-assisted review. These are suggestions only, so please verify before changing anything.

The fix for regular tasks is correct. The main change needed is in the recurring-task refresh (see the inline comment on `useRecurringTaskMutations.ts`).

blotztask-mobile/src/feature/calendar/hooks/useRecurringTaskMutations.ts
```diff
@@ -65,12 +82,9 @@ export function useRecurringTaskMutations() {
...
 }
 
-function invalidateRecurringOccurrenceQueries(
-  queryClient: QueryClient,
-  occurrenceDate: string,
-) {
+function invalidateRecurringOccurrenceQueries(queryClient: QueryClient, occurrenceDate: string) {
   const date = parseISO(occurrenceDate);
-  const dayKey = convertToDateTimeOffset(startOfDay(date));
+  const dayKey = format(date, "yyyy-MM-dd");
```
sol-wizard: This only refreshes the `yyyy-MM-dd` key for the day the occurrence starts, which misses two cases:
- **Multi-day recurring task viewed on a later day:** the checkbox stays stuck.
- **Home-screen widget:** it caches days under the `convertToDateTimeOffset` key, which the old code happened to match, so it no longer refreshes.

Suggest refreshing every selected-day cache instead:
```ts
queryClient.invalidateQueries({ queryKey: [...taskKeys.all, "selectedDay"] });
```

blotztask-mobile/src/feature/calendar/hooks/useRecurringTaskMutations.ts
```diff
@@ -30,6 +30,25 @@ export function useRecurringTaskMutations() {
   const { mutate: completeOccurrence, isPending: isCompletingOccurrence } = useMutation({
     mutationFn: ({ recurringTaskId, occurrenceDate }: CompleteRecurringOccurrenceArgs) =>
       saveRecurringOccurrence({ recurringTaskId, occurrenceDate }),
+    onMutate: (data) => {
```
sol-wizard, line 34: For the same multi-day case, `onMutate` updates the start day's cache, not the day being viewed. Could `task-card` pass `selectedDay` here, like `toggleTask` does?

blotztask-mobile/src/shared/hooks/useTaskMutations.ts
```diff
@@ -99,19 +99,20 @@ const useTaskMutations = () => {
 
   const toggleTaskMutation = useMutation({
     mutationFn: ({ taskId }: ToggleTaskVariables) => toggleTaskCompletion(taskId),
-    onMutate: async (data) => {
+    onMutate: (data) => {
```
sol-wizard, line 102: Suggest following the standard optimistic-update pattern:
- `await queryClient.cancelQueries(...)` at the start of `onMutate`, so a refetch from an earlier toggle can't overwrite this tick.
- Move the invalidate from `onSuccess` to `onSettled`. On a timeout where the server actually saved, the UI currently rolls back to "not done", and the next tap flips the task back to not done.

sol-wizard (approved):

PR #548, opened 2026-09-16
feat(review): lay out the monthly letter as theme, stats and suggestion

AlexisEvan:
## Summary

The monthly letter rendered as one block of prose even after the backend split it into parts. It now renders as separate blocks: the theme, the month's numbers, the body, and one thing to try next month.

- Theme gates the structured blocks — a weekly letter, a month too quiet to name a theme, and a letter written before the split all leave it null and render body-only. `tasksCompleted` is counted live and always has a value, so it cannot carry that signal
- The stats row renders whichever stats it was given rather than a fixed three: days active has no data source yet, focused minutes is deferred
- The weekly review is untouched — the new blocks are gated to monthly
- A quiet period now keeps its next-period suggestion instead of dropping it, so a low-activity month still closes on something to try

## Release note

> Your monthly review letter now shows the month's theme and numbers at a glance, and ends with one thing to try next month.

**Status:**


AlexisEvan:
截屏2026-09-17 上午12 16 04
截屏2026-09-17 上午12 15 50

AlexisEvan:
截屏2026-09-17 上午12 17 08

quiet month

blotztask-mobile/src/feature/review/components/letter-stats.tsx
```diff
@@ -0,0 +1,53 @@
...
+};
+
+// The design shows three stats, but only tasksCompleted has a data source today — so the row
+// renders whichever stats it was given rather than a fixed three.
+export function LetterStats({ tasksCompleted }: Props) {
+  const { t } = useTranslation("settings");
+
+  const stats: Stat[] = [];
+
+  if (tasksCompleted !== null) {
```
sol-wizard: A month with a theme but nothing ticked off shows a big green "0 tasks completed" — and it is inside the share card. The prompt tells the AI to be gentle about unfinished tasks, so this reads a bit harsh. Should we hide the stat when the count is 0?

blotztask-api/Modules/Reviews/Prompts/ReviewPrompts.cs
```diff
@@ -55,7 +55,7 @@ public static string GetReviewPrompt(
 
                 First, decide which tasks carry real signal about how the user actually spent or intended their time. Disregard entries that are not genuine activity — placeholder, sample or example content, tests, or text that reads as random rather than a real task. Base everything below only on the tasks that remain.
 
-                If little or no genuine activity remains after that, do NOT invent themes, patterns, or meaning. Put two or three warm, honest sentences acknowledging it was a quiet {periodNoun} in "body", leave "theme" and "oneThingToTryNext" null, and stop there.
+                If little or no genuine activity remains after that, do NOT invent themes, patterns, or meaning. Put two or three warm, honest sentences acknowledging it was a quiet {periodNoun} in "body" and leave "theme" null. Still write "oneThingToTryNext": for a quiet {periodNoun} it is not advice drawn from the data but one small, kind invitation for next {periodNoun} — something easy to say yes to, like noting one moment a day. Then stop there.
```
sol-wizard: With only one example ("noting one moment a day"), the model will likely copy it almost word for word, so a user with two quiet months in a row reads the same suggestion twice. Give two or three different examples, or add "vary it, do not reuse the example".

blotztask-mobile/src/feature/review/components/letter-card-content.tsx
```diff
@@ -49,9 +53,33 @@ export function LetterCardContent({
   } else if (isCurrentMonth) {
     content = <MonthlyLetterInProgressState />;
   } else if (report) {
+    // Theme is what marks a letter as written in parts — a quiet month and a pre-split letter
+    // both leave it null, and both render body-only. A quiet month still gets a suggestion.
+    const theme = periodType === ReviewPeriodType.Monthly ? report.theme : null;
```
sol-wizard, line 58: `!== null` lets `undefined` through. If the app reaches an API that does not send these fields yet (app released before the API), every monthly letter shows an empty theme block and an empty suggestion title. Use `!= null` here and on the line below — `useReview` already does that for `letter`.

blotztask-mobile/src/feature/review/models/review-dto.ts
```diff
@@ -9,7 +9,13 @@ export type ReviewReportDTO = {
   periodType: ReviewPeriodType;
   periodStartLocal: string;
   periodEndLocalExclusive: string;
+  // letter holds the body; theme and oneThingToTryNext are null on letters written before the
+  // backend split them out, and on a period too quiet to name a theme.
```
sol-wizard: nit: this is no longer true for `oneThingToTryNext` — the prompt change in this PR makes a quiet period still get a suggestion. Same stale wording in `ReviewLetter.cs:6`.

sol-wizard (approved):

PR #552, opened 2026-09-18
feat(onboarding): make the first page a try-voice step

sol-wizard:
## Summary

~31% of signed-in users never create a task or note, and onboarding only described voice. The first page now has the user hold the mic, say one thing to do today, and save the drafted task with "Add & continue", so Today is not empty on arrival.

- Reuses the AI sheet's flow: extracted `useSaveAiResults`, `useHoldToTalk`, `resolveMicPermission`; `useAiTaskGenerator` gains optional `onComplete`/`onError`. Sheet behaviour unchanged.
- `IntroCarousel`: swipe lock, per-item primary button, `keyboardShouldPersistTaps`.
- Never blocks: mic refused → typing; no speech, no task or network error → inline next step; Skip always works.
- Analytics: four `onboarding_voice_*` events, `task_created.source = "onboarding_ai"`.

Tested on a real iPhone: typed flow to a DB row and the Today list, hold-vs-swipe, Skip, server unresponsive, sheet regression. Not exercised: spoken audio, mic-refusal prompt.

## Release note

> New to Blotz? Onboarding now lets you create your first task by voice before you even reach the app.

**Status:**


sol-wizard:
Superseded by #554. Same goal, but it guides the user to the real AI sheet after onboarding instead of building a second voice UI inside the carousel.

PR #560, opened 2026-09-21
fix(ai-sheet): keep the sheet open when microphone access is off

AlexisEvan:
## Summary

The AI sheet asked for mic permission on mount and called `router.back()` when it was not
granted, so anyone who had once declined could never reach the sheet — or its keyboard input.
`permission/PermissionDenied` was the top AI failure in PostHog.

- Permission is requested on the first mic-button press, never on mount.
- The sheet opens in text mode unless the mic is already granted, and never closes itself.
- Declined or blocked shows an inline hint; blocked also gets an Open Settings button.
- Pressing the mic while blocked briefly swaps the hint to the reason — the OS shows no prompt
  in that state, so the press would otherwise look dead.
- Returning from Settings with the mic on switches back to voice input.
- Analytics keep the same outcomes, now reported once per sheet open.

## Release note

> The AI assistant now works without microphone access — it stays open so you can type your tasks instead.

**Status:**


AlexisEvan:
https://github.com/user-attachments/assets/0531e06b-3fd9-40cd-b235-abe86f08498c

Chinese Mode

AlexisEvan:
https://github.com/user-attachments/assets/4b6a693e-2976-4056-b354-16c52232e20b

English mode

dksjl661 (approved):

